PR #2, opened 2024-02-06
Benjamin sirianni

math-rad:
We might need to change some things before it should be merged.


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -71,4 +80,44 @@ public Command getAutonomousCommand() {
...
+  // 2020 IndexerSubsystem.java: https://github.com/STMARobotics/frc-7028-2020/blob/b4bafc1792f9b4ddc187f8470b0808450ce86903/src/main/java/frc/robot/subsystems/IndexerSubsystem.java#L90C3-L98C4 
+  // 2020 RobotContainer.java:   https://github.com/STMARobotics/frc-7028-2020/blob/b4bafc1792f9b4ddc187f8470b0808450ce86903/src/main/java/frc/robot/RobotContainer.java#L240C3-L300C4 
+  private void configureSmartDashboard() {
+    configureSubsystemTab();
+    configureDriverTab();
+    //configureSensorTab();
+  }
+
+  private void configureSubsystemTab() {
+    ShuffleboardLayout IndexerSubsystemLayout = subsystemTab.getLayout("Indexer", BuiltInLayouts.kList);
```
agasser: Variable names should start with lower-case (camel case)

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
 package frc.robot.subsystems;
 
 import static com.revrobotics.CANSparkLowLevel.MotorType.kBrushless;
-import static edu.wpi.first.units.Units.Volts;
+import edu.wpi.first.units.Units.Volts;
```
agasser: You need a static import for Volts.
Don't try to manage import manually, let VS Code do it for you.

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
...
-  private boolean isIntakeSensorTripped() {
-    return colorSensorReader.getIntakeValues().proximity > THRESHOLD_INTAKE;
+    indexerColorMatch.addColorMatch(COLOR_NOTE);
+    indexerColorMatch.addColorMatch(COLOR_NONE);
   }
-
-  public boolean isShooterSensorTripped() {
-    return colorSensorReader.getShooterValues().proximity > THRESHOLD_SHOOTER;
+  
+  public boolean shouldContinue() {
```
agasser: This is not a descriptive name. Should continue what?

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
...
+    return indexerColorMatch.matchClosestColor(indexerColorSensor.getColor()).color == COLOR_NOTE;
   }
 
-  public void stop() {
-    pidController.setReference(0, ControlType.kVelocity);
+  // Check actual motor values for dash board debugging
+  public boolean isActive() {
+    return leftIndexerMotor.get() > 0 || rightIndexerMotor.get() > 0;
```
agasser: I think `get()` only returns a non-zero value if you're running with duty cycle.
I'm pretty sure `getAppliedOutput()` or `getOutputCurrent()` will work with closed loop control, but you'll have to test it.

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
...
-  public IndexerSubsystem() {
-    indexerMotorConfig(leftIndexerMotor, true);
-    indexerMotorConfig(rightIndexerMotor, false);
 
-    colorSensorReader.run();
-    // Update the color sensors in the background to prevent loop overrun
-    colorSensorNotifier.setName("Color Sensors");
-    colorSensorNotifier.startPeriodic(0.02);
+  public void stopIndexer() {
+    leftPidController.setReference(0, ControlType.kVelocity);
```
agasser: I'm not sure we want to set the velocity to 0, vs. setting the output to 0. If the motors are running, this will apply negative output to stop the motors. This will also apply current to hold the velocity at zero if the motors try to spin. Although we _might_ need this to hold the note in, I'm pretty sure the compression will keep it secure without running the motors.

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
...
-    return colorSensorReader.getIntakeValues().proximity > THRESHOLD_INTAKE;
+    indexerColorMatch.addColorMatch(COLOR_NOTE);
+    indexerColorMatch.addColorMatch(COLOR_NONE);
   }
-
-  public boolean isShooterSensorTripped() {
-    return colorSensorReader.getShooterValues().proximity > THRESHOLD_SHOOTER;
+  
+  public boolean shouldContinue() {
+    return indexerColorMatch.matchClosestColor(indexerColorSensor.getColor()).color == COLOR_NOTE;
```
agasser: I don't think this logic is correct. Assuming "should continue" means there is _not_ a note in front of the sensor, this should be checking if the nearest match is `COLOR_NONE`, or _not_ `COLOR_NOTE`

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
...
   }
 
-  public void addDashboardWidgets(ShuffleboardLayout dashboard) {
-    var detailDashboard = dashboard.getLayout("Detail", BuiltInLayouts.kGrid)
-        .withProperties(Map.of("Number of columns", 2, "Number of rows", 3)).withPosition(0, 0);
-    detailDashboard.addNumber("Intake Prox", () -> colorSensorReader.getIntakeValues().proximity).withPosition(0, 0);
-    detailDashboard.addNumber("Shooter Prox", () -> colorSensorReader.getIntakeValues().proximity).withPosition(0, 0);
-    detailDashboard.addString("Ring Type", () -> this.getIndexerState().name()).withPosition(1, 1);
-
+  private SysIdRoutine indexerSysIdRoutine = new SysIdRoutine(
```
agasser: Fields should go before the constructor. This will compile and work, but it's confusing to humans. The order should be:
static fields
fields
constructors
methods (usually public then private, but not as important)

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -1,126 +1,133 @@
...
-    } else {
-      load();
-    }
+  // Driver tab 
+  public void addDriverDashboardWidgets(ShuffleboardTab driverDashboardTab) {
+    driverDashboardTab.addBoolean("Ring detected", this::shouldContinue);
+    driverDashboardTab.addBoolean("Active indexer motors", this::isActive);
   }
 
+  // Semioperable implementation of smart dashboard, edited by me before I did a small revamp above
```
agasser: Please don't commit code that is commented out.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -71,4 +80,44 @@ public Command getAutonomousCommand() {
     return autoChooser.getSelected();
   }
 
+  // Ben. S, 2/4/2024 1:16AM
+  // Remove if unnecessary, or edit parts that aren't correct or lack needed info - I still tried my best 😭
+
+  // Shuffleboard is comprised of tabs, which then *can* contain widgets OR layouts made up of widgets
```
agasser: FYI, tabs are layouts. Layouts can contain more layouts, or widgets.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -71,4 +80,44 @@ public Command getAutonomousCommand() {
...
+  //| Smart dashboard
+  //|       Tab:                          Generally made up of layouts or their own independent widgets
+  //|          -> Widget:                 GUI for information, sometimes interactable
+  //|          -> Layout:                 They are essentially just subfolders for widgets used for organization
+  //|                   -> Widget
+  // 
+  // https://docs.wpilib.org/en/stable/docs/software/dashboards/shuffleboard/layouts-with-code/organizing-widgets.html#adding-widgets-to-layouts  
+  //  
+  // 2020 IndexerSubsystem.java: https://github.com/STMARobotics/frc-7028-2020/blob/b4bafc1792f9b4ddc187f8470b0808450ce86903/src/main/java/frc/robot/subsystems/IndexerSubsystem.java#L90C3-L98C4 
+  // 2020 RobotContainer.java:   https://github.com/STMARobotics/frc-7028-2020/blob/b4bafc1792f9b4ddc187f8470b0808450ce86903/src/main/java/frc/robot/RobotContainer.java#L240C3-L300C4
```
agasser: I don't think this comment belongs in the code. It doesn't explain anything that's specific to this codebase, the information is all part of the standard WPILib documentation.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -71,4 +80,44 @@ public Command getAutonomousCommand() {
...
+  //|                   -> Widget
+  // 
+  // https://docs.wpilib.org/en/stable/docs/software/dashboards/shuffleboard/layouts-with-code/organizing-widgets.html#adding-widgets-to-layouts  
+  //  
+  // 2020 IndexerSubsystem.java: https://github.com/STMARobotics/frc-7028-2020/blob/b4bafc1792f9b4ddc187f8470b0808450ce86903/src/main/java/frc/robot/subsystems/IndexerSubsystem.java#L90C3-L98C4 
+  // 2020 RobotContainer.java:   https://github.com/STMARobotics/frc-7028-2020/blob/b4bafc1792f9b4ddc187f8470b0808450ce86903/src/main/java/frc/robot/RobotContainer.java#L240C3-L300C4 
+  private void configureSmartDashboard() {
+    configureSubsystemTab();
+    configureDriverTab();
+    //configureSensorTab();
```
agasser: Please remove commented code

agasser (changes requested):
I have a few comments. I understand you are learning this, so I wouldn't have expected you to know these things!

PR #3, opened 2024-02-08
Robot 1 - For review only

agasser:
This is a PR for reviewing Robot 1. It won't be merged to `main`.


src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
-    public static final double FIELD_WIDTH_METERS = 8.0137;
+  }
 
-    /** Minimum target ambiguity. Targets with higher ambiguity will be discarded */
-    public static final double APRILTAG_AMBIGUITY_THRESHOLD = 0.2;
+  public static class ShooterConstants {
+    public static final int DEVICE_ID_SHOOTER_RIGHT = 51;
+    public static final int DEVICE_ID_SHOOTER_LEFT = 50;
+    public static final int DEVICE_ID_ACTUATOR_MOTOR = 52;
+    public static final int SHOOTER_VELOCITY_OFFSET = 0;
```
agasser: The naming doesn't seem right on this. I think it's a tolerance, not an offset.

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+    public static final int WRIST_POSITION_OFFSET = 0;
```
agasser: Two problems here:
1) This constant is _used_ as a tolerance in the subsystem, not an offset.
2) There needs to be an offset configured in the subsystem, but it is not.

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+    public static final int SHOOTER_MOTOR_RATIO = 1;
```
agasser, line 116: This is unused

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+
+    public static final SlotConfigs SHOOTER_SLOT_CONFIGS = new SlotConfigs()
+        .withKP(0.59)
+        .withKI(0)
+        .withKD(0.001)
+        .withKS(0);
+  }
+
+  public static class ClimbConstants {
+    public static DeadbandFilter CLIMB_DEADBAND_FILTER = new DeadbandFilter(-.25, .5);
```
agasser: This is unused

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+
+    public static final int DEVICE_ID_FIRST_STAGE_CLIMB = 12;
```
agasser, line 133: This is unused

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+
+    public static final double SOFT_LIMIT_FIRST_STAGE_FWD = 430000;
+    public static final double SOFT_LIMIT_FIRST_STAGE_REV = 0;
```
agasser, line 136: These are not used, but they should be. They should be `Measure<Length>` or `Measure<Rotations>`

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+        .withKP(0.59)
+        .withKI(0)
+        .withKD(0.001)
+        .withKS(0);
+  }
+
+  public static class ElevatorConstants {
+    public static final int ELEVATOR_LEADER_ID = 60;
+    public static final int ELEVATOR_FOLLOWER_ID = 61;
+    public static final int ANALOG_SENSOR_CHANNEL = 0;
```
agasser, line 173: This is unused

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+  public static class IndexerConstants {
+    public static final int PORT_ID_INTAKE_SENSOR = 3;
+    public static final int PORT_ID_SHOOTER_SENSOR = 2;
+
+    public static final int DEVICE_ID_INDEXER_LEFT_MOTOR = 1;
+    public static final int DEVICE_ID_INDEXER_RIGHT_MOTOR = 2;
+
+    // Proximity thresholds for when to trip each sensor
+    public static final int THRESHOLD_INTAKE = 240;
+    public static final int THRESHOLD_SHOOTER = 225;
```
agasser, line 205: These are unused. Should they be used?

src/main/java/frc/robot/Constants.java
```diff
@@ -58,18 +63,110 @@ public static class AutoDriveConstants {
...
+
+    public static final double BELT_kP = 0d;
+    public static final double BELT_RUN_SPEED = 8000;
+    public static final double BELT_SHOOT_SPEED = 8000;
```
agasser, line 209: There is no "belt", so these should be renamed (or deleted if they aren't used)

src/main/java/frc/robot/DeadbandFilter.java
```diff
@@ -0,0 +1,55 @@
+package frc.robot;
+
+/**
+ * A filter that can handle a deadband, useful for joysticks that don't center
+ * at exactly 0.
+ */
+public class DeadbandFilter {
```
agasser: This is now built in to WPILib, we don't need to have our own any more. The method is `MathUtil.applyDeadband()`

src/main/java/frc/robot/Mechanisms.java
```diff
@@ -0,0 +1,62 @@
...
+import edu.wpi.first.wpilibj.smartdashboard.MechanismLigament2d;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.wpilibj.util.Color;
+import edu.wpi.first.wpilibj.util.Color8Bit;
+
+/**
+ * Class to keep all the mechanism-specific objects together and out of the main
+ * example
+ */
+public class Mechanisms {
```
agasser: I don't think this class / file belongs in our code. Why is it here?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -46,29 +58,93 @@ public RobotContainer() {
...
-    controlBindings.wheelsToX().ifPresent(trigger -> trigger.whileTrue(drivetrain.applyRequest(() -> brake)));
-    controlBindings.resetPose().ifPresent(trigger -> trigger.onTrue(drivetrain.runOnce(drivetrain::seedFieldRelative)));
+
+    // drive
+    controlBindings.wheelsToX().ifPresent(trigger -> trigger.whileTrue(driveTrain.applyRequest(() -> brake)));
+    controlBindings.resetPose().ifPresent(trigger -> trigger.onTrue(driveTrain.runOnce(driveTrain::seedFieldRelative)));
+
+    // shooter
+    controlBindings.shootDutyCycle().ifPresent(trigger -> trigger.whileTrue(Commands.startEnd(
+        () -> shooterSubsystem.shootDutyCycle(2), shooterSubsystem::stop, shooterSubsystem)));
```
agasser: Duty cycle is a value in the range [-1, 1]. This is essentially telling the shooter to run at 200%.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -46,29 +58,93 @@ public RobotContainer() {
...
+
+    // indexer
+    controlBindings.indexerRun().ifPresent(trigger -> trigger.whileTrue(Commands.startEnd(
+        () -> indexerSubsystem.fireDonut(Volts.of(3)), indexerSubsystem::stop, indexerSubsystem)));
```
agasser: Is this right that `indexerRun()` is bound to `fireDonut`? By the names, I would expect `indexerRun` to run the indexer, probably to intake a note, so it would stop when the note gets to the color sensor. I would expect `fireDonut` to send the donut into the shooter.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -46,29 +58,93 @@ public RobotContainer() {
...
+    tab.add("Drive Steer Dynam F", driveTrain.sysIdSteerDynamCommand(kForward)).withPosition(4, 1);
+    tab.add("Drive Steer Quasi R", driveTrain.sysIdSteerQuasiCommand(kReverse)).withPosition(4, 2);
+    tab.add("Drive Steer Dynam R", driveTrain.sysIdSteerDynamCommand(kReverse)).withPosition(4, 3);
+
+    tab.add("Drive Drive Quasi F", driveTrain.sysIdDriveQuasiCommand(kForward)).withPosition(5, 0);
+    tab.add("Drive Drive Dynam F", driveTrain.sysIdDriveDynamCommand(kForward)).withPosition(5, 1);
+    tab.add("Drive Drive Quasi R", driveTrain.sysIdDriveQuasiCommand(kReverse)).withPosition(5, 2);
+    tab.add("Drive Drive Dynam R", driveTrain.sysIdDriveDynamCommand(kReverse)).withPosition(5, 3);
+
+    tab.add("Drive Slip F", driveTrain.sysIdDriveSlipCommand()).withPosition(7, 0);
```
agasser, line 136: It'll work, but it's a little confusing that column 7 comes before 6 in the code.

src/main/java/frc/robot/commands/ColorSensorReader.java
```diff
@@ -0,0 +1,53 @@
...
+import frc.robot.util.MultiplexedColorSensor;
+
+/**
+ * Runnable to use with the WPILib Notifier to update the color sensors in the background.
+ * This class enables color sensors to be read on a background thread to avoid overrunning
+ * the main robot loop.
+ */
+public class ColorSensorReader implements Runnable {
+  
+  private final MultiplexedColorSensor intakeColorSensor = new MultiplexedColorSensor(Port.kMXP, PORT_ID_INTAKE_SENSOR);
```
agasser: Robot 1 only has one color sensor, and it is not connected to a multiplexer.

src/main/java/frc/robot/commands/DeployIntakeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.robot.subsystems.IntakeSubsystem;
+
+/**
+ * Command to run shooter motors
+ */
+public class DeployIntakeCommand extends Command {
+
+  private static final double DEPLOY_TIME = 0.25;
```
agasser: You can use a `Timer`, it's not _wrong_, but a more elegant way to handle simple timing is with the `withTimeout()`. You don't need to add your own logic to the command.
Something like:
```
var commandWithTimeout = new DeployIntakeCommand(60, intakeSubsystem).withTimeout(0.25);
```

See https://docs.wpilib.org/en/stable/docs/software/commandbased/command-compositions.html#adding-command-end-conditions

src/main/java/frc/robot/commands/DeployIntakeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+  private final IntakeSubsystem intakeSubsystem;
+  private final Timer deployTimer = new Timer();
+  protected double shooterRPS = 1;
```
agasser: This should be `private final`, unless there is a specific reason to expose to a larger scope and make it mutable.

src/main/java/frc/robot/commands/DeployIntakeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public DeployIntakeCommand(double shooterRPS, IntakeSubsystem intakeSubsystem) {
+    this.intakeSubsystem = intakeSubsystem;
```
agasser, line 15: The constructor takes `shooterRPS` but doesn't do anything with it. I would guess it's supposed to be setting `this.shooterRPS`

src/main/java/frc/robot/commands/DeployIntakeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+  @Override
+  public void execute() {
+    intakeSubsystem.deploy(Volts.of(3));
+    intakeSubsystem.intakeRollers(Volts.of(3));
+    deployTimer.start();
+  }
+
+  @Override
+  public boolean isFinished() {
+    return deployTimer.hasElapsed(DEPLOY_TIME);
```
agasser: Checking the position of the intake would be more reliable than time.

src/main/java/frc/robot/commands/RunShooterMotor.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Command to run shooter motors
+ */
+public class RunShooterMotor extends Command {
+
+  private static final double SHOOT_TIME = 0.25;
+  private final ShooterSubsystem shooterSubsystem;
+  private final Timer shootTimer = new Timer();
+  protected double shooterRPS = 1;
```
agasser: This should be `private final`, unless there is a specific reason to expose to a larger scope and make it mutable.

src/main/java/frc/robot/commands/DeployIntakeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+public class DeployIntakeCommand extends Command {
+
+  private static final double DEPLOY_TIME = 0.25;
+  private final IntakeSubsystem intakeSubsystem;
+  private final Timer deployTimer = new Timer();
+  protected double shooterRPS = 1;
+
+  public DeployIntakeCommand(double shooterRPS, IntakeSubsystem intakeSubsystem) {
+    this.intakeSubsystem = intakeSubsystem;
+  }
```
agasser, line 19: We **must** call `addRequirements(intakeSubsystem)`

https://docs.wpilib.org/en/stable/docs/software/commandbased/commands.html#getrequirements

src/main/java/frc/robot/commands/RunShooterMotor.java
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Constructor
+   * 
+   * @param shooterRPS       velocity of shooter in rotations per second
+   * @param shooterSubsystem shooter
+   */
+  public RunShooterMotor(double shooterRPS, ShooterSubsystem shooterSubsystem) {
+    this.shooterRPS = shooterRPS;
+    this.shooterSubsystem = shooterSubsystem;
+  }
```
agasser: We **must** call `addRequirements(shooterSubsystem)`

https://docs.wpilib.org/en/stable/docs/software/commandbased/commands.html#getrequirements

src/main/java/frc/robot/commands/RunShooterMotor.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public void initialize() {
+    shooterSubsystem.activeStop();
```
agasser: Why is this here?

src/main/java/frc/robot/commands/RunShooterMotor.java
```diff
@@ -0,0 +1,50 @@
...
+import frc.robot.subsystems.ShooterSubsystem;
+
+/**
+ * Command to run shooter motors
+ */
+public class RunShooterMotor extends Command {
+
+  private static final double SHOOT_TIME = 0.25;
+  private final ShooterSubsystem shooterSubsystem;
+  private final Timer shootTimer = new Timer();
```
agasser: Same as above, a more elegant way to handle simple timing is using `withTimeout()` to wrap the command.

src/main/java/frc/robot/commands/RunShooterMotor.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Override
+  public boolean isFinished() {
+    return shootTimer.hasElapsed(SHOOT_TIME);
+  }
+
+  @Override
+  public void end(boolean interrupted) {
+    shooterSubsystem.actuatorStop();
```
agasser: Why stop the actuator when the command didn't do anything else with it?

src/main/java/frc/robot/commands/ShootDonutCommand.java
```diff
@@ -0,0 +1,75 @@
...
+public class ShootDonutCommand extends Command {
+
+  private static final double SHOOT_TIME = 0.25;
+  private final ShooterSubsystem shooterSubsystem;
+  private final IndexerSubsystem indexerSubsystem;
+  private final VelocityAngleInterpolator velocityAngleInterpolator;
+  private final Supplier<Pose2d> robotPoseSupplier;
+
+  private final Timer shootTimer = new Timer();
+  protected double shooterRPS = 1;
```
agasser: This should be `private final`, unless there is a specific reason to expose to a larger scope and make it mutable.

src/main/java/frc/robot/commands/RunShooterMotor.java
```diff
@@ -0,0 +1,50 @@
...
+   */
+  public RunShooterMotor(double shooterRPS, ShooterSubsystem shooterSubsystem) {
+    this.shooterRPS = shooterRPS;
+    this.shooterSubsystem = shooterSubsystem;
+  }
+
+  @Override
+  public void initialize() {
+    shooterSubsystem.activeStop();
+  }
```
agasser: If you keep the timer, you should reset it here so the command can be scheduled more than once.

src/main/java/frc/robot/commands/DeployIntakeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Override
+  public void initialize() {
```
agasser, line 22: If you keep the timer, you should reset it here so the command can be scheduled more than once.

src/main/java/frc/robot/commands/SpitCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+import edu.wpi.first.wpilibj.Timer;
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.robot.subsystems.IntakeSubsystem;
+
+public class SpitCommand extends Command {
+
+  private static final double SHOOT_TIME = 0.25;
+  private final IntakeSubsystem intakeSubsystem;
+  private final Timer shootTimer = new Timer();
```
agasser: Same comment as above about using `withTimeout`.

src/main/java/frc/robot/commands/SpitCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public SpitCommand(IntakeSubsystem intakeSubsystem) {
+    this.intakeSubsystem = intakeSubsystem;
+  }
+
+  @Override
+  public void initialize() {
+  }
```
agasser: If you keep the timer, you should reset it here

src/main/java/frc/robot/commands/SpitCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class SpitCommand extends Command {
+
+  private static final double SHOOT_TIME = 0.25;
+  private final IntakeSubsystem intakeSubsystem;
+  private final Timer shootTimer = new Timer();
+
+  public SpitCommand(IntakeSubsystem intakeSubsystem) {
+    this.intakeSubsystem = intakeSubsystem;
+  }
```
agasser: We **must** call `addRequirements(intakeSubsystem)`

src/main/java/frc/robot/controls/ControlBindings.java
```diff
@@ -10,8 +10,18 @@
 import edu.wpi.first.wpilibj2.command.button.Trigger;
 
 public interface ControlBindings {
+  Optional<Trigger> elevatorVelocity();
```
agasser: This name isn't clear. Does it make the elevator go up or down?

src/main/java/frc/robot/controls/ControlBindings.java
```diff
@@ -10,8 +10,18 @@
...
   Optional<Trigger> resetPose();
   Optional<Trigger> wheelsToX();
+  Optional<Trigger> shootDutyCycle();
+  Optional<Trigger> spit();
```
agasser: Spit what from where? I have a feeling this is reversing the intake and indexer to eject a note, so maybe `ejectNote`, or `reverseIntake` would be more clear.

src/main/java/frc/robot/controls/ControlBindings.java
```diff
@@ -10,8 +10,18 @@
...
+  Optional<Trigger> intakeRollers();
```
agasser: This should be a verb (action word), not a noun (thing). So probably something like `runIntakeRollers` or `intakeNote`

src/main/java/frc/robot/controls/ControlBindings.java
```diff
@@ -10,8 +10,18 @@
...
+  Optional<Trigger> deployIntake();
+  Optional<Trigger> retractIntake();
+  Optional<Trigger> indexerRun();
+  Optional<Trigger> actuatorUp();
+  Optional<Trigger> actuatorDown();
+  Optional<Trigger> runShooterMotor();
```
agasser: This seems a lot like `shootDutyCycle`, it might make sense to have names that explain the difference.

src/main/java/frc/robot/sim/PhysicsSim.java
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.ArrayList;
+
+import com.ctre.phoenix6.Utils;
+import com.ctre.phoenix6.hardware.TalonFX;
+
+/**
+ * Manages physics simulation for CTRE products.
+ */
+public class PhysicsSim {
```
agasser, line 11: Why is this class / file here?

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ClimbSubsystem extends SubsystemBase {
+
+  private final TalonFX climbConfig = new TalonFX(DEVICE_ID_FIRST_STAGE_CLIMB);
+
+  public ClimbSubsystem() {
+
+    climbConfig.setSafetyEnabled(true);
+    climbConfig.setNeutralMode(NeutralModeValue.Brake);
+    climbConfig.getConfigurator().apply(new TalonFXConfiguration());
```
agasser, line 21: This line will reset the settings to factory default, so line 20 would be reversed. You probably should set the neutral mode in the `TalonFXConfiguration`.

```
TalonFXConfiguration config = new TalonFXConfiguration();
config.MotorOutput.NeutralMode = NeutralModeValue.Brake;

climbConfig.getConfigurator().apply(config));
```

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+import com.ctre.phoenix6.hardware.TalonFX;
+import com.ctre.phoenix6.signals.NeutralModeValue;
+
+import edu.wpi.first.wpilibj.shuffleboard.BuiltInWidgets;
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardTab;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class ClimbSubsystem extends SubsystemBase {
+
+  private final TalonFX climbConfig = new TalonFX(DEVICE_ID_FIRST_STAGE_CLIMB);
```
agasser, line 15: This name is confusing, it's not a config, it's a motor controller.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public void stopFirstStage() {
+    climbConfig.stopMotor();
+  }
+
+  public boolean isFirstStageRaised() {
+    var rotorPosSignal = climbConfig.getRotorPosition();
+    var rotorPos = rotorPosSignal.getValue();
+    assert rotorPos > 3000;
```
agasser, line 36: Don't use `assert`, it will throw an exception and crash the robot program if the condition is not true.

https://www.baeldung.com/java-assert

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+    return true;
```
agasser, line 37: This logic isn't right. It will either crash the robot program, or return `true`

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.ctre.phoenix6.configs.TalonFXConfiguration;
+import com.ctre.phoenix6.hardware.TalonFX;
+import com.ctre.phoenix6.signals.NeutralModeValue;
+
+import edu.wpi.first.wpilibj.shuffleboard.BuiltInWidgets;
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardTab;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class ClimbSubsystem extends SubsystemBase {
```
agasser, line 13: Robot 1 has no climb. Maybe we should just delete this

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+  private static final double MOTOR_BOTTOM = 0;
+  private static final double MOTOR_TOP = 56530;
+
+  // Mutiply by sensor position to get meters
+  private static final double MOTOR_ENCODER_POSITION_COEFFICIENT = ELEVATOR_HEIGHT / (MOTOR_TOP - MOTOR_BOTTOM);
+
+  private static final int ANALOG_BOTTOM = 758;
+  private static final int ANALOG_TOP = 1796;
+
+  private static final double GRAVITY_FEED_FORWARD = 0.05;
```
agasser, line 59: We should probably put all of these in `Constants` (if they're used)

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+
+  // Limit switches - FALSE means at limit
+  private final DigitalInput topLimitSwitch = new DigitalInput(8);
+  private final DigitalInput bottomLimitSwitch = new DigitalInput(9);
+
+  private double targetPosition = 0;
+
+  public ElevatorSubsystem() {
+    elevatorLeader.getConfigurator().apply(new TalonFXConfiguration());
+    elevatorFollower.getConfigurator().apply(new TalonFXConfiguration());
```
agasser, line 71: This will restore the configuration to factory defaults. I don't think that was the intention.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+    double kV = 0.12;
+    double kIz = 0;
+    double kS = 0.25;
+    double kMaxOutput = 0.7;
+    double kMinOutput = -.4;
+    double allowedErr = 1;
+
+    // Magic Motion Coefficients
+    double maxVel = 10000;
+    double maxAcc = 30000;
```
agasser, line 88: These aren't used.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+
+    TalonFXConfiguration config = new TalonFXConfiguration();
+    config.Slot0 = Slot0Configs.from(ELEVATOR_SLOT_CONFIGS);
+    config.MotorOutput.NeutralMode = NeutralModeValue.Brake;
+    config.MotorOutput.Inverted = InvertedValue.CounterClockwise_Positive;
+    elevatorLeader.getConfigurator().apply(config);
+    elevatorFollower.getConfigurator().apply(config);
+
+    elevatorLeader.getConfigurator().refresh(config);
+    elevatorFollower.getConfigurator().refresh(config);
```
agasser, line 98: The `refresh()` method will load the motor's configuration into the config object. Is there a reason for doing this?

https://pro.docs.ctr-electronics.com/en/latest/docs/api-reference/api-usage/configuration.html

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  private static double kDt = 0.02;
+
+  // Create a motion profile with the given maximum velocity and maximum
+  // acceleration constraints for the next setpoint.
+  private final TrapezoidProfile m_profile = new TrapezoidProfile(new TrapezoidProfile.Constraints(1.75, 0.75));
+  private TrapezoidProfile.State m_goal = new TrapezoidProfile.State();
+  private TrapezoidProfile.State m_setpoint = new TrapezoidProfile.State();
+  private TrapezoidProfile.State m_feedforward = new TrapezoidProfile.State();
```
agasser, line 190: 1) Fields should go at the top. This really confused me when I was looking for where they were defined.
2) We aren't putting `m_` in front of our field names, we're using camelCase
3) I don't think any of this stuff is needed because we should be using [`MotionMagicVoltage`](https://pro.docs.ctr-electronics.com/en/latest/docs/api-reference/device-specific/talonfx/motion-magic.html) on the motor controller.

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  public void moveElevator(double speed) {
+    targetPosition = 0;
+    elevatorLeader.set(speed);
+  }
+
+  public void moveToPosition(double meters) {
+    targetPosition = meters;
+    elevatorLeader.set(meters);
```
agasser, line 125: There's a unit error here. Presumably, the `meters` parameter is in meters, but `set` takes a percentage in the range of [-1, 1]

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  public Command sysIdElevatorQuasiCommand(Direction direction) {
+    return elavatorSysidRoutine.quasistatic(direction).withName("SysId Shooter Motors Quasistatic " + direction)
```
agasser, line 135: The name is wrong

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+        .finallyDo(this::stop);
+  }
+
+  public Command sysIdElevatorDynamCommand(Direction direction) {
+    return elavatorSysidRoutine.dynamic(direction).withName("SysId Shooter Motors Quasistatic " + direction)
```
agasser, line 140: The name is wrong

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+        .finallyDo(this::stop);
+  }
+
+  public double getElevatorPosition() {
+    var rotorPosSignal = elevatorLeader.getRotorPosition();
```
agasser, line 145: You probably don't want the rotor position. Given that you're using `setPosition` in `periodic`, I think you want `getPosition()`

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+    var rotorPos = rotorPosSignal.getValue();
+    rotorPosSignal.waitForUpdate(0.020);
```
agasser, line 147: Why wait for an updated value here, especially if we're not going to use it?

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+    return rotorPos;
```
agasser, line 148: Is it useful to the caller to get the elevator position in rotations of the motor? I think this should probably return meters (ideally, `Measure<Distance>`)

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  public double getElevatorTopPosition() {
+    return getElevatorPosition() * .489;
```
agasser, line 152: What is `.489`?

src/main/java/frc/robot/subsystems/ElevatorSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+    return !topLimitSwitch.get();
+  }
+
+  static double motorPositionToMeters(double motorPosition) {
+    return (motorPosition * MOTOR_ENCODER_POSITION_COEFFICIENT);
+  }
+
+  static double metersToMotorPosition(double positionMeters) {
+    return (positionMeters / MOTOR_ENCODER_POSITION_COEFFICIENT);
+  }
```
agasser, line 181: 1) These aren't used. Should they be?
2) Default visibility (package private) is almost certainly not the correct choice. Should these be `private`? 
http://xahlee.info/java-a-day/access_specifiers.html

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+  private static SparkPIDController indexerMotorConfig(CANSparkMax sparkMax, boolean invert) {
+    SparkPIDController pidController = sparkMax.getPIDController();
+    sparkMax.restoreFactoryDefaults();
+    sparkMax.enableVoltageCompensation(12);
+    sparkMax.setOpenLoopRampRate(0.1);
+    sparkMax.setClosedLoopRampRate(0.1);
+    sparkMax.setInverted(invert);
+    pidController.setP(IndexerConstants.BELT_kP);
+    pidController.setFF(0.00009);
+    sparkMax.getEncoder();
```
agasser: The return from `getEncoder()` isn't used. Why is this call being made?

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+
+  public IndexerSubsystem() {
+    leftPidController = indexerMotorConfig(leftIndexerMotor, false);
+    rightPidController = indexerMotorConfig(rightIndexerMotor, true);
+
+    indexerColorMatch.addColorMatch(COLOR_NOTE);
+    indexerColorMatch.addColorMatch(COLOR_NONE);
+  }
+
+  public boolean shouldContinue() {
```
agasser: This name is not descriptive. Should continue what?

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+    return indexerColorMatch.matchClosestColor(indexerColorSensor.getColor()).color == COLOR_NOTE;
```
agasser: Assuming "should continue" means, should continue intaking, I think this logic is backwards. Don't you want to stop the indexer when the sensor sees a note, not run when it sees a note?

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  public boolean isActive() {
+    return leftIndexerMotor.get() > 0 || rightIndexerMotor.get() > 0;
```
agasser, line 61: I commented on this before. Check the API, I don't think `get()` will return a non-zero value when running closed-loop.

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  public void runIndexer() {
+    leftPidController.setReference(BELT_RUN_SPEED, ControlType.kVelocity);
+    rightPidController.setReference(BELT_RUN_SPEED, ControlType.kVelocity);
+  }
+
+  public void stopIndexer() {
+    leftPidController.setReference(0, ControlType.kVelocity);
+    rightPidController.setReference(0, ControlType.kVelocity);
```
agasser: I also commented on this before. I think we want to set the motors to idle, _not_ set the motors to closed-loop velocity zero. I think you should delete this method and use the other `stop()` method that sets the voltage to zero that's already implemented on this class.

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+
+  private SysIdRoutine indexerSysIdRoutine = new SysIdRoutine(
+      new SysIdRoutine.Config(null, null, null, SysIdRoutineSignalLogger.logState()),
+      new SysIdRoutine.Mechanism((volts) -> {
+        leftIndexerMotor.setVoltage(volts.in(Volts));
+        rightIndexerMotor.setVoltage(volts.in(Volts));
+      }, null, this));
+
+  public Command sysIdIndexerMotorQuasiCommand(Direction direction) {
+    return indexerSysIdRoutine.quasistatic(direction).withName("SysId Deploy Motor Quasistatic " + direction)
```
agasser, line 92: The name is wrong

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+        .finallyDo(this::stopIndexer);
+  }
+
+  public Command sysIdIndexerMotorDynamCommand(Direction direction) {
+    return indexerSysIdRoutine.dynamic(direction).withName("SysId Deploy Motor Quasistatic " + direction)
```
agasser, line 97: The name is wrong

src/main/java/frc/robot/subsystems/IndexerSubsystem.java
```diff
@@ -0,0 +1,126 @@
...
+
+  // Subsystem tab
+  public void addSubsystemDashboardWidgets(ShuffleboardLayout indexerLayout) {
+    indexerLayout.addBoolean("Ring detected", this::shouldContinue);
+    indexerLayout.addBoolean("Active motors", this::isActive);
+  }
+
+  // Driver tab
+  public void addDriverDashboardWidgets(ShuffleboardTab driverDashboardTab) {
+    driverDashboardTab.addBoolean("Ring detected", this::shouldContinue);
```
agasser: Let's call it a "note", of if we're being goofy, a "donut".

agasser: I'm pretty sure `shouldContinue` should return false when a note is detected, so this looks wrong.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+import edu.wpi.first.units.Voltage;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine;
+import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine.Direction;
+import frc.robot.subsystems.sysid.SysIdRoutineSignalLogger;
+
+public class IntakeSubsystem extends SubsystemBase {
+
+  private VoltageOut voltageRequest = new VoltageOut(0);
```
agasser, line 47: I think we want to enable FOC, unless there's a reason not to
https://pro.docs.ctr-electronics.com/en/latest/docs/api-reference/device-specific/talonfx/talonfx-control-intro.html#field-oriented-control

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+  private final TalonFX deployMotor = new TalonFX(DEVICE_ID_DEPLOY_MOTOR, CANIVORE_BUS_NAME);
+  private final TalonFX RollersMotor = new TalonFX(DEVICE_ID_ROLLERS_MOTOR, CANIVORE_BUS_NAME);
```
agasser: Field names should be camelCase (not start with an upper case R)

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
agasser, line 47: This can be `final`

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    intakeRollersConfig.Slot0 = Slot0Configs.from(ROLLERS_SLOT_CONFIGS);
+    intakeRollersConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    RollersMotor.getConfigurator().apply(intakeRollersConfig);
+    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
+    deployMotor.getConfigurator().apply(intakeDeployConfig);
+    CANcoderConfiguration canCoderConfig = new CANcoderConfiguration();
+    canCoderConfig.MagnetSensor.AbsoluteSensorRange = AbsoluteSensorRangeValue.Signed_PlusMinusHalf;
+    canCoderConfig.MagnetSensor.SensorDirection = SensorDirectionValue.CounterClockwise_Positive;
+    canCoderConfig.MagnetSensor.MagnetOffset = 0.4;
+    canCoder.getConfigurator().apply(canCoderConfig);
```
agasser: This is kind of mixed up, can we organize it better? Just changing the order and adding some meaningful blank lines can help a lot with clarity
```java
    var intakeRollersConfig = new TalonFXConfiguration();
    intakeRollersConfig.Slot0 = Slot0Configs.from(ROLLERS_SLOT_CONFIGS);
    intakeRollersConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
    RollersMotor.getConfigurator().apply(intakeRollersConfig);

    var intakeDeployConfig = new TalonFXConfiguration();
    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
    deployMotor.getConfigurator().apply(intakeDeployConfig);
    
    CANcoderConfiguration canCoderConfig = new CANcoderConfiguration();
    canCoderConfig.MagnetSensor.AbsoluteSensorRange = AbsoluteSensorRangeValue.Signed_PlusMinusHalf;
    canCoderConfig.MagnetSensor.SensorDirection = SensorDirectionValue.CounterClockwise_Positive;
    canCoderConfig.MagnetSensor.MagnetOffset = 0.4;
    canCoder.getConfigurator().apply(canCoderConfig);
```

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    var intakeDeployConfig = new TalonFXConfiguration();
+    intakeRollersConfig.Slot0 = Slot0Configs.from(ROLLERS_SLOT_CONFIGS);
+    intakeRollersConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    RollersMotor.getConfigurator().apply(intakeRollersConfig);
+    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
+    deployMotor.getConfigurator().apply(intakeDeployConfig);
+    CANcoderConfiguration canCoderConfig = new CANcoderConfiguration();
+    canCoderConfig.MagnetSensor.AbsoluteSensorRange = AbsoluteSensorRangeValue.Signed_PlusMinusHalf;
+    canCoderConfig.MagnetSensor.SensorDirection = SensorDirectionValue.CounterClockwise_Positive;
+    canCoderConfig.MagnetSensor.MagnetOffset = 0.4;
```
agasser: This value should be in `Constants`

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    var intakeRollersConfig = new TalonFXConfiguration();
+    var intakeDeployConfig = new TalonFXConfiguration();
+    intakeRollersConfig.Slot0 = Slot0Configs.from(ROLLERS_SLOT_CONFIGS);
+    intakeRollersConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    RollersMotor.getConfigurator().apply(intakeRollersConfig);
+    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
+    deployMotor.getConfigurator().apply(intakeDeployConfig);
+    CANcoderConfiguration canCoderConfig = new CANcoderConfiguration();
+    canCoderConfig.MagnetSensor.AbsoluteSensorRange = AbsoluteSensorRangeValue.Signed_PlusMinusHalf;
+    canCoderConfig.MagnetSensor.SensorDirection = SensorDirectionValue.CounterClockwise_Positive;
```
agasser: This should be `Clockwise_Positive`. It has to do with the way the sensor is physically mounted on the intake.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+          this));
+
+  public IntakeSubsystem() {
+    var intakeRollersConfig = new TalonFXConfiguration();
+    var intakeDeployConfig = new TalonFXConfiguration();
+    intakeRollersConfig.Slot0 = Slot0Configs.from(ROLLERS_SLOT_CONFIGS);
+    intakeRollersConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    RollersMotor.getConfigurator().apply(intakeRollersConfig);
+    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
+    deployMotor.getConfigurator().apply(intakeDeployConfig);
```
agasser, line 80: The deploy motor should be inverted. It has to do with the way the sensor is physically mounted on the intake. This way UP will be positive, and DOWN will be negative output.
`intakeDeployConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;`

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    canCoder.getConfigurator().apply(canCoderConfig);
+  }
+
+  private final VelocityTorqueCurrentFOC IntakeRollersMotorVelocity = new VelocityTorqueCurrentFOC(0, 0, 0, 1, false,
+      false,
+      false);
+
+  private final VelocityTorqueCurrentFOC IntakeDeployMotorVelocity = new VelocityTorqueCurrentFOC(0, 0, 0, 1, false,
+      false,
+      false);
```
agasser: 1) Fields should be before constructor
2) Fields should be camelCase

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Example command factory method.
+   *
+   * @return a command
+   */
```
agasser, line 107: This comment is wrong

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+  public void intakeRollers(Measure<Voltage> volts) {
+    RollersMotor.setControl(voltageRequest.withOutput(volts.in(Volts)));
+  }
+
+  public void deploy(Measure<Voltage> volts) {
+    var intakeDeployConfig = new TalonFXConfiguration();
+    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
+    intakeDeployConfig.MotorOutput.Inverted = InvertedValue.CounterClockwise_Positive;
+    deployMotor.getConfigurator().apply(intakeDeployConfig);
```
agasser: We shouldn't reconfigure the invert on the fly like this. Instead, just apply negative voltage to run the mechanism downward.
1) It takes a moment to apply, so it might cause loop overruns.
2) The motor will now be out of phase with the cancoder (one gets larger while the other gets smaller)
3) If we want to use arm gravity feed forward (which we will want to do), motion has to be upward positive.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    deployMotor.setControl(voltageRequest.withOutput(volts.in(Volts)));
+  }
+
+  public void retractIntake(Measure<Voltage> volts) {
+    var intakeDeployConfig = new TalonFXConfiguration();
+    intakeDeployConfig.Slot0 = Slot0Configs.from(DEPLOY_SLOT_CONFIGS);
+    intakeDeployConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    deployMotor.getConfigurator().apply(intakeDeployConfig);
```
agasser: Same comment as above about configuring on the fly, just apply positive voltage to run the intake upward.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    deployMotor.setControl(voltageRequest.withOutput(volts.in(Volts)));
+  }
+
+  // Spit means to reverse the motor direction to 'spit' the game piece out
```
agasser, line 112: Instead of a comment to describe the meaning of the method name, just use a more descriptive name. Java method names can be up to 65535 characters (obviously try to keep them under 30ish)

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+    return rollersMotorSysIdRoutine.dynamic(direction).withName("SysId Rollers Motor Quasistatic " + direction)
+        .finallyDo(this::stop);
+  }
+
+  public void stop() {
+    // READ THIS!!!!!
+    // IDK what the motor velocity number should actually be, just set it to 0 for
+    // now
+    deployMotor.setControl(IntakeDeployMotorVelocity.withVelocity(0));
+    RollersMotor.setControl(IntakeRollersMotorVelocity.withVelocity(0));
```
agasser: Don't set the velocity to zero to stop the motors. This will actively apply opposing voltage to the motor to stop it and hold it at a velocity of zero. We just want to apply idle output, so use `stopMotor()`.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,157 @@
...
+  // }
+  //
+  // @Override
+  // public void periodic() {
+  // // This method will be called once per scheduler run
+  // }
+  //
+  // @Override
+  // public void simulationPeriodic() {
+  // // This method will be called once per scheduler run during simulation
```
agasser, line 182: Remove these comments. They shouldn't be in git

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  private final TalonFX shooterRightMotor = new TalonFX(DEVICE_ID_SHOOTER_RIGHT, CANIVORE_BUS_NAME);
+  private final CANSparkMax actuatorMotor = new CANSparkMax(DEVICE_ID_ACTUATOR_MOTOR, MotorType.kBrushless);
+  // creates new devices and sets them to the device Ids set in
+  // Constants.java
+
+  private static final double ENCODER_OFFSET = -0.2285f;
+  private static final float LIMIT_BOTTOM = 0.12f;
+  private static final float LIMIT_TOP = 0.4272f;
+  private static final double LIMIT_TOP_RADIANS = Units.rotationsToRadians(LIMIT_TOP + ENCODER_OFFSET);
+  private static final double LIMIT_BOTTOM_RADIANS = Units.rotationsToRadians(LIMIT_BOTTOM + ENCODER_OFFSET);
```
agasser, line 55: These should be in `Constants`, and they should be reviewed.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  private boolean isActiveStopped = false;
```
agasser, line 56: I don't think we need an active stop mode, do we? That was for hanging on to cones in Orien, I'm not sure that really applies here.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  public boolean hasRing = false;
```
agasser, line 57: This isn't used.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  private SparkAbsoluteEncoder actuatorEncoder;
```
agasser, line 58: This can be `final`

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  private final SparkPIDController pidController;
+  private VoltageOut voltageRequest = new VoltageOut(0).withEnableFOC(true);
```
agasser, line 60: This can be `final`

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private SysIdRoutine shooterMotorSysIdRoutine = new SysIdRoutine(
+      new SysIdRoutine.Config(null, null, null, SysIdRoutineSignalLogger.logState()),
+      new SysIdRoutine.Mechanism((volts) -> {
+        shooterRightMotor.setControl(voltageRequest.withOutput(volts.in(Volts)));
+        shooterLeftMotor.setControl(voltageRequest.withOutput(volts.in(Volts)));
+      }, null, this));
+
+  private final VelocityTorqueCurrentFOC shooterMotorVelocity = new VelocityTorqueCurrentFOC(0, 0, 0, 1, false, false,
```
agasser, line 72: We'll also need sysID routines for the wrist/aim. Since it's a REV motor, we can use URCL. See https://docs.wpilib.org/en/stable/docs/software/advanced-controls/system-identification/creating-routine.html#routine-config

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+    shooterMotorConfig.OpenLoopRamps.DutyCycleOpenLoopRampPeriod = 1;
+    shooterMotorConfig.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    shooterLeftMotor.getConfigurator().apply(shooterMotorConfig);
+    shooterMotorConfig.MotorOutput.Inverted = InvertedValue.CounterClockwise_Positive;
+    shooterRightMotor.getConfigurator().apply(shooterMotorConfig);
+
+    // Get the through-bore-encoder absolute encoder
+    actuatorEncoder = actuatorMotor.getAbsoluteEncoder(kDutyCycle);
+    actuatorEncoder.setInverted(true);
+    actuatorEncoder.setAverageDepth(64);
```
agasser, line 89: I think you should set the encoder offset, instead of manually applying it later. This will also allow the soft limits to work.
`actuatorEncoder.setZeroOffset(AIM_OFFSET.in(Rotations));`

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+    pidController = actuatorMotor.getPIDController();
+    pidController.setFeedbackDevice(actuatorEncoder);
+
```
agasser: Actuator motor configuration should happen here, only once.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  }
```
agasser, line 92: The actuator motor should have soft limits configured
```java
    aimMotor.setSoftLimit(kForward, AIM_FORWARD_LIMIT);
    aimMotor.setSoftLimit(kReverse, AIM_REVERSE_LIMIT);
    aimMotor.enableSoftLimit(kForward, true);
    aimMotor.enableSoftLimit(kReverse, true);
```

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private static SparkPIDController indexerMotorConfig(CANSparkMax sparkMax, boolean invert) {
```
agasser: This looks like it was copied from the indexer. It should be reviewed, and a separate method isn't really needed since there is only one Spark Max controller.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+   * Gets the wrist position Zero is horizontal, up is positive
+   * 
+   * @return position in radians
+   */
+  public double getWristPosition() {
+    return Units.rotationsToRadians(actuatorEncoder.getPosition() + ENCODER_OFFSET);
+  }
+
+  public void actuatorUp(Measure<Voltage> volts) {
+    indexerMotorConfig(actuatorMotor, true);
```
agasser: Controllers should not be configured on the fly. Just apply positive output to go up.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+    actuatorMotor.setVoltage(volts.in(Volts));
+  }
+
+  public void actuatorDown(Measure<Voltage> volts) {
+    indexerMotorConfig(actuatorMotor, false);
```
agasser: Controllers should not be configured on the fly. Just apply negative output to go down.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+    shooterRightMotor.setControl(shooterMotorVelocity.withVelocity(shooterRPS));
+  }
+
+  public Command sysIdShooterMotorQuasiCommand(Direction direction) {
+    return shooterMotorSysIdRoutine.quasistatic(direction).withName("SysId Shooter Motors Quasistatic " + direction)
+        .finallyDo(this::stop);
+  }
+
+  public Command sysIdShooterMotorDynamCommand(Direction direction) {
+    return shooterMotorSysIdRoutine.dynamic(direction).withName("SysId Shooter Motors Quasistatic " + direction)
```
agasser, line 121: The name is wrong

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+        .finallyDo(this::stop);
+  }
+
+  public void activeStop() {
+    if (!isActiveStopped) {
+      shooterLeftMotor.setControl(shooterMotorVelocity.withVelocity(0));
+      shooterRightMotor.setControl(shooterMotorVelocity.withVelocity(0));
+    }
+    isActiveStopped = true;
+  }
```
agasser, line 148: Do we need an active stop?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Gets the wrist position Zero is horizontal, up is positive
+   * 
+   * @return position in radians
+   */
+  public double getWristPosition() {
+    return Units.rotationsToRadians(actuatorEncoder.getPosition() + ENCODER_OFFSET);
```
agasser, line 156: It would make sense for the encoder to have the offset configured, instead of applying it here.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+    indexerMotorConfig(actuatorMotor, false);
+    actuatorMotor.setVoltage(volts.in(Volts));
+  }
+
+  /**
+   * Gets the wrist velocity in radians per second
+   * 
+   * @return wrist velocity in radians per second
+   */
+  public double getWristVelocity() {
```
agasser, line 174: I guess this is fine, but why would we use it?

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+    return Units.rotationsToDegrees(actuatorEncoder.getVelocity());
+  }
+
+  public boolean checkWristPosition(double radiansToRotate) {
+    return (Math.abs(getWristPosition() - radiansToRotate) <= WRIST_POSITION_OFFSET);
+  }
+
+  public boolean checkShooterSpeed(double shooterSpeedGoal) {
+    return (Math.abs(getWristVelocity() - shooterSpeedGoal) <= SHOOTER_VELOCITY_OFFSET);
```
agasser: This logic is not correct. It's checking the velocity of the wrist, not the velocity of the shooter.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Gets the wrist velocity in radians per second
+   * 
+   * @return wrist velocity in radians per second
+   */
+  public double getWristVelocity() {
+    return Units.rotationsToDegrees(actuatorEncoder.getVelocity());
+  }
+
+  public boolean checkWristPosition(double radiansToRotate) {
```
agasser: I think it would be more useful to have a method that returns `true` if the wrist is at the previously requested velocity. Having it this way will be hard to deal with in a command. The same applies to the `checkShooterSpeed` method.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,181 @@
...
+  public void stop() {
+    shooterRightMotor.stopMotor();
+    shooterLeftMotor.stopMotor();
+    isActiveStopped = false;
+
+  }
+
+  public void actuatorRotate(double radians) {
+    var target = MathUtil.clamp(radians, LIMIT_BOTTOM_RADIANS, LIMIT_TOP_RADIANS);
+    new TrapezoidProfile.State(target, 0.0);
```
agasser, line 214: This method doesn't actually do anything

src/main/java/frc/robot/util/MultiplexedColorSensor.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * This is a wrapper for the REV Color Sensor V3, allowing multiple sensors to be
+ * used with a multiplexer board (this was tested using the Adafruit TCA9548A).
+ * 
+ * This class only exposes a single {@link #getValues()} method to get both the color
+ * and proximity. For our purposes, we only need these two values, and we want both of
+ * them for every robot loop. Combining them into a single method allows us to minimize
+ * the writes to switch ports on the multiplexer.
+ */
+public class MultiplexedColorSensor {
```
agasser: We're not going to have color sensors on a multiplexer

agasser (commented):
One comment overall, let's consider taking advantage of the new [Java Units Library](https://docs.wpilib.org/en/stable/docs/software/basic-programming/java-units.html) as much as we can, to avoid unit errors.

PR #5, opened 2024-02-22
Robot 2 auto

finnsherman:
Adds all stuff needed for 6 and 7 note autonomous


finnsherman:
This code is probably wrong somehow but I am going to create a pull request so Andy can tell me how its wrong.

src/main/java/frc/robot/math/VelocityPitchYawInterpolator.java
agasser, line 1: Something isn't right with this branch, this file shouldn't be included, it already exists in `main`.

src/main/java/frc/robot/Constants.java
agasser, line 1: Please revert the reformatting of this file, it got weird.

src/main/java/frc/robot/Constants.java
```diff
@@ -272,9 +282,12 @@ public static class ShootingConstants {
...
 
-    public static final VelocityPitchInterpolator SHOOTER_INTERPOLATOR = new VelocityPitchInterpolator(List.of(
-      new ShootingSettings().distance(Meters.of(1)).velocity(RotationsPerSecond.of(40)).pitch(Rotations.of(0.1))
-    ));
+    public static final VelocityPitchInterpolator SHOOTER_PITCH_INTERPOLATOR = new VelocityPitchInterpolator(List.of(
+        new ShootingSettings().distance(Meters.of(1)).velocity(RotationsPerSecond.of(40)).pitch(Rotations.of(0.1))));
+
+    public static final VelocityPitchYawInterpolator SHOOTER_PITCH_YAW_INTERPOLATOR = new VelocityPitchYawInterpolator(
+        List.of(new ShootingSettingsYaw().distance(Meters.of(1)).velocity(RotationsPerSecond.of(40)).pitch(
+            Rotations.of(0.1)).yaw(Rotations.of(0.1))));
```
agasser: The `VelocityPitchYawInterpolator` does pitch and yaw, that's its only reason for existing. There should only be one table.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -73,6 +76,10 @@ public RobotContainer() {
       controlBindings = new JoystickControlBindings();
     }
 
+    NamedCommands.registerCommand("aimAndShoot", autoBuilder.aimAndShoot());
+    NamedCommands.registerCommand("shootDonut", autoBuilder.shootDonut());
+    NamedCommands.registerCommand("startIntake", autoBuilder.startIntake());
```
agasser: I haven't thought too much about this, but it seems like this probably belongs in AutonomousBuilder, I think named commands are its purpose.

src/main/java/frc/robot/commands/AutoAimShootCommand.java
agasser, line 1: How is this different from `ShootCommand`?

src/main/java/frc/robot/commands/AutoStartIntakeCommand.java
agasser, line 1: This could just be an `InstantCommand` . Is there a reason we need it in a separate command like this?

src/main/java/frc/robot/commands/ShootCommand.java
```diff
@@ -74,7 +74,7 @@ public void execute() {
     var distanceToSpeaker = robotTranslation.getDistance(speakerTranslation);
 
     // Lookup shooter settings for this distance
-    var shootingSettings = SHOOTER_INTERPOLATOR.calculate(distanceToSpeaker);
+    var shootingSettings = SHOOTER_PITCH_INTERPOLATOR.calculate(distanceToSpeaker);
```
agasser: Why is this changed? The interpolator does both pitch and velocity, why put only pitch in the name?

agasser:
Thanks for opening this PR early for comment. I didn't look at the paths/autos yet, we should try them (probably next week) before merging to main.

Please don't reformat entire files. It introduces a bunch of changes and makes it hard to find what you actually changed. The formatter also isn't very smart, and sometimes is wraps lines in ways that make the code hard to read. It also increases the odds of getting merge conflicts, since it's changing so much.

agasser (changes requested):

src/main/java/frc/robot/RobotContainer.java
finnsherman: PathPlanner documentation tells me it should be in robot container, I moved it anyway but I'm not 100% sure it will work.

https://pathplanner.dev/pplib-named-commands.html

src/main/java/frc/robot/commands/AutoStartIntakeCommand.java
agasser: I did a little more thinking about this, and I couldn't come up with a reason why you would want to only run the intake. What's the purpose of this command?

src/main/java/frc/robot/commands/AutoAimShootCommand.java
finnsherman: `ShootCommand` aims the drivetrain, `AutoAimShootCommand` aims the turret.

src/main/java/frc/robot/RobotContainer.java
agasser: OK, yeah I see that, it says it's recommended to register them in RobotContainer. PP's example code is mostly using [instance factory methods](https://docs.wpilib.org/en/stable/docs/software/commandbased/organizing-command-based.html#instance-command-factory-methods). I messed around with this in my robot 1 code, but I found it to be restrictive since we have so many commands that span subsystems.

I'm open to discussing this, but since you already created a [non-static command factory](https://docs.wpilib.org/en/stable/docs/software/commandbased/organizing-command-based.html#non-static-command-factories), and that has worked well for us in the past, I thought we would probably continue with that design. If we do, I think it makes sense to have the factory responsible for registering the commands. This way, the addition or removal of a command only requires a change in one class.

Something that programmers talk a lot about is ["The Single Responsibility Principle"](https://en.wikipedia.org/wiki/Single_responsibility_principle) which means "Gather together the things that change for the same reasons. Separate those things that change for different reasons."

src/main/java/frc/robot/Constants.java
finnsherman: I made `VelocityPitchYawInterpolator` in order to be able to calculate Yaw, as horizontal aiming is controlled by the drivetrain in many places, however for auto I would like it to just be the Turret aiming. I'm pretty sure having that all in one table would make the calculations incorrect for many of your manual shooting commands, as they would take into account Yaw when they don't use it. If you do make `VelocityPitchInterpolator` calculate yaw at some point, and update the Commands to rotate the turret then we should merge them, however I feel it is best to keep them in seperate tables for now.

However I am probably missing something here, so feel free to correct me.

src/main/java/frc/robot/commands/AutoAimShootCommand.java
agasser: The turret's yaw defines the aim of the shooter. Yaw needs to be set based on the location of the target, not a lookup table.

image

`ShootCommand` (now renamed `ScoreSpeakerCommand`) will be responsible for the drivetrain heading (yaw) and the turret yaw, we just haven't written that yet. Like the 2022 robot, the goal will be to score by only turning the turret, unless the speaker is out of reach of the turret, then the drivetrain heading will be adjusted so the turret can reach. You can watch it work in [this youtube video](https://youtu.be/oHHk32bZ5hI?si=BsajTSs45Hm8uA2o&t=51). Onyx is in the top right corner with red bumpers. See how the driver approaches the hub with the intake forward, but the robot spins around, turns the turret and scores.

The _ultimate_ goal would be if you guys can calculate what's need to shoot while moving. The driver would maintain translation control, but the heading (yaw) would be set by the code. An offset of distance and angle would be calculated to compensate for the robot's motion.

src/main/java/frc/robot/Constants.java
agasser: See my other comment that I added while you were typing this.

I see that I misunderstood and didn't think deeply enough on this yesterday. My comment at the top of this thread is confused. `VelocityPitchInterpolator` is used to look up the velocity of the shooter wheels and the pitch of the turret. The yaw of the turret (and the drivetrain) should be defined by the location of the target (and optionally, the speed of the robot.) Yaw shouldn't be in a lookup table.

src/main/java/frc/robot/AutonomousBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+      TurretSubsystem turretSubsystem,
+      CommandSwerveDrivetrain driveTrain) {
+    this.intakeSubsystem = intakeSubsystem;
+    this.shooterSubsystem = shooterSubsystem;
+    this.turretSubsystem = turretSubsystem;
+    this.driveTrain = driveTrain;
+
+  NamedCommands.registerCommand("aimAndShoot", aimAndShoot());
+  NamedCommands.registerCommand("shootDonut", shootDonut());
+  NamedCommands.registerCommand("startIntake", startIntake());
```
agasser, line 33: These lines aren't indented right

src/main/java/frc/robot/AutonomousBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+  
+  }
+
+  public Pose2d poseSupplier() {
+    return driveTrain.getState().Pose;
+  }
+
+  public Command aimAndShoot() {
+    return new AutoAimShootCommand(shooterSubsystem, turretSubsystem, poseSupplier());
```
agasser, line 42: This is passing the pose, not the `poseSupplier()` method reference. I think you want `this::poseSupplier`

src/main/java/frc/robot/AutonomousBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  public Command startIntake() {
+    return Commands.startEnd(() -> intakeSubsystem.intake(), null, intakeSubsystem);
```
agasser, line 46: I'm not sure you can use `StartEndCommand` with a `null` end `Runnable`. But, why not use use `InstantCommand` with `Commands.runOne`?

src/main/java/frc/robot/AutonomousBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public Pose2d poseSupplier() {
+    return driveTrain.getState().Pose;
+  }
+
+  public Command aimAndShoot() {
+    return new AutoAimShootCommand(shooterSubsystem, turretSubsystem, poseSupplier());
+  }
+
+  public Command startIntake() {
```
agasser, line 45: Why would we ever want to start only the intake rollers?

src/main/java/frc/robot/AutonomousBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+    return Commands.startEnd(() -> intakeSubsystem.intake(), null, intakeSubsystem);
+  }
+
+  public Command shootDonut() {
+    return new ManualShootCommand(turretSubsystem, shooterSubsystem);
```
agasser, line 50: Are you sure `ManualShootCommand` is what you'd want to use? It shoots "point blank" at a fixed velocity, pitch, and yaw.

src/main/java/frc/robot/commands/AutoAimShootCommand.java
agasser, line 1: Is this supposed to do something different than `ScoreSpeakerCommand`?

src/main/java/frc/robot/commands/AutoAimShootCommand.java
```diff
@@ -0,0 +1,93 @@
...
+  private final Pose2d poseSupplier;
+
+  private final Timer shootTimer = new Timer();
+
+  private Translation2d speakerTranslation;
+
+  private final MutableMeasure<Angle> angleToRotate = MutableMeasure.zero(Rotations);
+
+  public AutoAimShootCommand(ShooterSubsystem shooter,
+  TurretSubsystem turretSubsystem, Pose2d poseSupplier) {
```
agasser, line 36: Indenting is wrong

agasser, line 36: Your argument `poseSupplier` is just a pose, not a supplier. I think you need a `Supplier<Pose>` so you can get the current post, not just a single pose at the time the object is constructed.

src/main/java/frc/robot/commands/AutoAimShootCommand.java
```diff
@@ -0,0 +1,93 @@
...
+
+    // Lookup shooter settings for this distance
+    var shootingSettings = SHOOTER_INTERPOLATOR.calculate(distanceToSpeaker);
+
+    // Prepare shooter
+    shooter.prepareToShoot(shootingSettings.getVelocity());
+
+    // Prepare turret
+    turretSubsystem.moveToPitchPosition(shootingSettings.getPitch());
+    turretSubsystem.moveToYawShootingPosition(angleToRotate);
```
agasser, line 72: This will be wrong by 180-degrees

src/main/java/frc/robot/commands/AutoAimShootCommand.java
```diff
@@ -0,0 +1,93 @@
...
+    var alliance = DriverStation.getAlliance();
+    speakerTranslation = (alliance.isEmpty() || alliance.get() == Blue) ? SPEAKER_BLUE : SPEAKER_RED;
+  }
+
+  @Override
+  public void execute() {
+    var robotPose = poseSupplier;
+    var robotTranslation = robotPose.getTranslation();
+
+    // Angle to turn the robot. The shooter is on the back, so it's the angle to the speaker plus PI radians.
```
agasser, line 57: This comment _would_ be right, but your code isn't turning the robot.

src/main/java/frc/robot/math/MobileShootCalculator.java
agasser, line 1: What is this class? It kind of looks like it's reproducing some of the things `Translation2d` can do.
This class is never used.

src/main/java/frc/robot/math/MobileShootCalculator.java
```diff
@@ -0,0 +1,34 @@
...
+
+import edu.wpi.first.math.geometry.Translation2d;
+import edu.wpi.first.math.kinematics.ChassisSpeeds;
+
+public class MobileShootCalculator {
+  
+  private final ChassisSpeeds chassisSpeeds = new ChassisSpeeds();
+  ArrayList<Double> currentSpeeds = new ArrayList<Double>();
+
+  public MobileShootCalculator() {
```
agasser, line 13: I don't think this class would work. The constructor doesn't set `currentSpeeds`. `currentSpeeds` is package-private (default visibility), so it seems like the intent might have been that a caller (in the same package) would directly set the field, but they wouldn't be able to do that before constructing the object. In short, you'd always end up with zero.

agasser (changes requested):

src/main/java/frc/robot/AutonomousBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.Commands;
+import frc.robot.commands.AutoAimShootCommand;
+import frc.robot.commands.ManualShootCommand;
+import frc.robot.subsystems.CommandSwerveDrivetrain;
+import frc.robot.subsystems.IntakeSubsystem;
+import frc.robot.subsystems.ShooterSubsystem;
+import frc.robot.subsystems.TurretSubsystem;
+
+public class AutonomousBuilder {
```
agasser, line 15: No instances of this object are ever created, so the `NamedCommands` won't get registered. We need to create an instance of this class before calling `AutoBuilder.buildAutoChooser();`

src/main/java/frc/robot/AutonomousBuilder.java
agasser, line 1: This should be merged with `AutoCommands` or removed.

agasser (changes requested):

PR #7, opened 2024-03-01
added limelight subsystem and integrated commands

finnsherman:
limelight subsystem and commands that integrate it, I'm like 98% sure I did this wrong.


agasser:
I thought we weren't going to use LimelightCalcs. It was intended to solve problems we don't have on this robot.

src/main/java/frc/robot/Robot.java
```diff
@@ -21,6 +22,10 @@ public void robotInit() {
     m_robotContainer = new RobotContainer();
 
     SignalLogger.start(); // CTRE logger
+
+    for (int port = 5800; port <= 5807; port++) {
```
agasser: Those ports are already forwarded to a different host.
https://github.com/STMARobotics/frc-7028-2024/blob/main/src/main/java/frc/robot/PhotonRunnable.java#L62

I also don't think `Robot` it the right place to do this in a command-based robot.

src/main/java/frc/robot/commands/AutoIntakeToAmperCommand.java
agasser, line 1: This class needs to be complete or removed before merging to `main`.

src/main/java/frc/robot/commands/AutoIntakeToTurretCommand.java
agasser, line 1: This class needs to be complete or removed before merging to `main`.

src/main/java/frc/robot/commands/IntakeToAmperCommand.java
agasser, line 1: Were these changes supposed to be in `AutoIntakeToAmperCommand`?

src/main/java/frc/robot/commands/IntakeToAmperCommand.java
```diff
@@ -11,16 +39,82 @@ public class IntakeToAmperCommand extends Command {
...
+  IntakeSubsystem intakeSubsystem,  
+  AmperSubsystem amperSubsystem,
+  CommandSwerveDrivetrain drivetrainSubsystem,
+  LimelightSubsystem limelightSubsystem, 
+  Supplier<Measure<Velocity<Distance>>> translationXSupplier,
+  Supplier<Measure<Velocity<Distance>>> translationYSupplier,
+  Supplier<Measure<Velocity<Angle>>> rotationSupplier,
+  LimelightResults limelightResults,
+  LimelightCalcs limelightCalcs,
+  LimelightDetectorTarget limelightDetectorTarget) {
```
agasser: The parameters aren't indented properly

src/main/java/frc/robot/commands/IntakeToAmperCommand.java
```diff
@@ -11,16 +39,82 @@ public class IntakeToAmperCommand extends Command {
...
   @Override
   public void execute() {
+    desiredChassisSpeeds.vxMetersPerSecond = translationXSupplier.get().in(MetersPerSecond);
+    desiredChassisSpeeds.vyMetersPerSecond = translationYSupplier.get().in(MetersPerSecond);
+    desiredChassisSpeeds.omegaRadiansPerSecond = rotationSupplier.get().in(RadiansPerSecond);
+    var limitedCassisSpeeds = rateLimiter.calculate(desiredChassisSpeeds);
+    drivetrainSubsystem.setControl(drive
+        .withVelocityX(limitedCassisSpeeds.vxMetersPerSecond)
+        .withVelocityY(limitedCassisSpeeds.vyMetersPerSecond)
+        .withRotationalRate(limitedCassisSpeeds.omegaRadiansPerSecond));
```
agasser: I think you only want to let the driver have full control if there has never been a valid target

src/main/java/frc/robot/commands/IntakeToAmperCommand.java
```diff
@@ -11,16 +39,82 @@ public class IntakeToAmperCommand extends Command {
...
+    if (limelightResults.valid == true) {
```
agasser: This is OK as is, but I have two tips:
1) There is no need to use `== true`, `valid` is already a boolean, so you can just do `if (limelightResults.valid)`
2) It's a best practice in cases like this (if you were to keep the `== true`) to put the literal on the left. It's a bit quirky, but it will help you catch a mistake at compile time instead of runtime. Consider this mistakes:
```java
if (limelightResults.valid = true) { // always true
  System.out.println("true");
}
```
This code will always print `true`. The bug is that the _assignment_ `=` was used instead of _equality check_ `==`. This will assign `limelightresults.valid` equal to `true` and return the value `true`, so the condition is also always `true`.

If we flip the same mistake around, we'll get a compile error because you cannot assign a value to the literal `true`:
```java
if (true = limelightResults.valid ) {  // compile error
  System.out.println("true");
}
```
 
Of these two options, I prefer number 1

src/main/java/frc/robot/commands/IntakeToAmperCommand.java
```diff
@@ -11,16 +39,82 @@ public class IntakeToAmperCommand extends Command {
...
+      desiredChassisSpeeds.vxMetersPerSecond = translationXSupplier.get().in(MetersPerSecond);
+      desiredChassisSpeeds.vyMetersPerSecond = translationYSupplier.get().in(MetersPerSecond);
+      desiredChassisSpeeds.omegaRadiansPerSecond = getTargetRotation().getRadians();
```
agasser: I think we need to use a PIDController (or preferably, a ProfiledPIDController) here. We don't just want to rotate at a speed equal to the target's angle.

src/main/java/frc/robot/commands/IntakeToTurretCommand.java
agasser, line 1: Let's one of these intake commands working, and then we'll see if we can fine a way to avoid duplicating all of the code into another command.

src/main/java/frc/robot/limelight/LimelightCalcs.java
agasser, line 1: Are you sure we want to use this?

src/main/java/frc/robot/limelight/LimelightClassifierTarget.java
agasser, line 1: Let's use [limelight-lib](https://github.com/LimelightVision/limelightlib-wpijava) this year. It was still a work in progress last year, so we just brought in parts. It would replace the rest of the files in this PR.

agasser (changes requested):

src/main/java/frc/robot/AutoCommands.java
```diff
@@ -35,17 +35,25 @@ public class AutoCommands {
...
-      LEDSubsystem ledSubsystem, ElevatorSubsystem elevatorSubsystem) {
+  public AutoCommands(
+    AmperSubsystem amperSubsystem, 
+    CommandSwerveDrivetrain drivetrainSubsystem,
+    ShooterSubsystem shooterSubsystem, 
+    TurretSubsystem turretSubsystem, 
+    IntakeSubsystem intakeSubsystem,
+    LEDSubsystem ledSubsystem, 
+    ElevatorSubsystem elevatorSubsystem,
+    LimelightHelpers limelightHelpers) {
```
agasser: Still indented wrong

src/main/java/frc/robot/subsystems/LimelightSubsystem.java
agasser, line 1: Are you going to keep LimelightSubsystem?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -58,12 +60,13 @@ public class RobotContainer {
   private final TurretSubsystem turretSubsystem = new TurretSubsystem();
   private final ClimbSubsystem climbSubsystem = new ClimbSubsystem();
   private final LEDSubsystem ledSubsystem = new LEDSubsystem();
+  private final LimelightHelpers limelightHelpers = new LimelightHelpers();
```
agasser: LimelightHelpers is all static. We will never instantiate it

agasser (changes requested):

agasser:
Replaced by #10

PR #9, opened 2024-03-01
Shoot while move code V1

dylanrattai:


src/main/java/frc/robot/commands/ScoreSpeakerMovingCommand.java
```diff
@@ -0,0 +1,217 @@
...
+    var robotTranslation = robotPose.getTranslation();
+
+    // Distance between the robot and the speaker
+    var distanceToSpeaker = robotTranslation.getDistance(speakerTranslation);
+
+    // Lookup shooter settings for this distance
+    var shootingSettings = SHOOTER_INTERPOLATOR.calculate(distanceToSpeaker);
+
+    // Calculate time to hit speaker
+    var timeUntilScored = SHOOTER_COEFFICIENT * distanceToSpeaker * SHOOTER_INTERPOLATOR.calculate(distanceToSpeaker).getVelocity().in(RotationsPerSecond);
```
agasser: `shootSettings` already has the calculated value, you don't need to call `SHOOTER_INTERPOLATOR.calculate(distanceToSpeaker)` again

src/main/java/frc/robot/commands/ScoreSpeakerMovingCommand.java
```diff
@@ -0,0 +1,217 @@
...
+
+    // Calculate the predicted offset of the speaker compared to current pose (in meters)
+    var speakerPredictedOffset = new Translation2d((drivetrain.getState().speeds.vxMetersPerSecond * timeUntilScored),
```
agasser: I think you want to use `driveTrain.getCurrentFieldChassisSpeeds()` because you want the field relative chassis speeds.

agasser (changes requested):

PR #11, opened 2024-03-02
Test mode

finnsherman:
test mode for pit, I am actually 100% sure there is a better way to do this, specifically,

-Not calling UpdatedProgressBarCommand periodically in Robot.java
-Not having an entirely separate file for test mode

Also I'm pretty sure the LEDProgressBarCommand doesn't really work either.


agasser:
If you're only interacting with the LEDs, you can try all of your code in the simulation.
Columns: 4
Order: Column Major
Start: Lower Left
Serpentine: true
image

src/main/java/frc/robot/commands/led/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @Override
+  public boolean runsWhenDisabled() {
+    return true;
+  }
+
+  @Override
+  public void end(boolean interrupted) {
+  }
```
agasser: Call ledSubsystem.setUpdater(null);

src/main/java/frc/robot/commands/led/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,71 @@
...
+    
+  @Override
+  public boolean isFinished() {
+    return done;
+  }
+
+  @Override
+  public boolean runsWhenDisabled() {
+    return true;
```
agasser: You can remove this whole method override, because the command will finish if you're not in test mode, so there's no reason to run when disabled.

src/main/java/frc/robot/commands/led/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private void animate(LEDStrips ledStrips) {
+      for(int strip = 0; strip < ledStrips.getStripCount(); strip++) {
+        for(index = 0; index < tests*2; index++) {
+            ledStrips.setLED(strip, index, Color.kBlue);
+        }
+      }
+      ledStrips.refresh();
+      done = !RobotState.isTest();
```
agasser: You can just put `return !RobotState.isTest();` and get rid of the `done` variable

src/main/java/frc/robot/commands/led/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public void UpdateTestingDashboard() {
+    
+    tests = testsCompleted.getDouble(0);
+  }
+
+  private void animate(LEDStrips ledStrips) {
+      for(int strip = 0; strip < ledStrips.getStripCount(); strip++) {
+        for(index = 0; index < tests*2; index++) {
+            ledStrips.setLED(strip, index, Color.kBlue);
```
agasser: If you want to, you can set the other LEDs to black, so they go off.

agasser (changes requested):

finnsherman:
So I believe that I have successfully made a Command that runs all needed tests in sequential order, while updating the LEDs with the number of tests. It builds, and there are no errors when running it in the simulator.

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,59 @@
...
+      for(index = 0; index < ledStrips.getStripSize(); index++) { 
+        if (index<tests*2) {
+          ledStrips.setLED(strip, index, Color.kBlue);
+        } else {
+          ledStrips.setLED(strip, index, Color.kBlack);
+        }
+      }
+    }
+    ledStrips.refresh();
+    }
```
agasser: Indenting isn't right

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -276,7 +279,14 @@ public void populateSysIdDashboard() {
...
     tab.add("Tur Roll Dynam Fwd", turretSubsystem.sysIdRollerDynamicCommand(kForward)).withPosition(columnIndex, 2);
     tab.add("Tur Roll Dynam Rev", turretSubsystem.sysIdRollerDynamicCommand(kReverse)).withPosition(columnIndex, 3);
-
+  
+    // Testing
+    tab = Shuffleboard.getTab("Testing");
+    
+    tab.addNumber("Number of Tests Run", () -> testCommand.getTestState());
+    tab.add("Start LED Progress Bar", new LEDProgressBarCommand(ledSubsystem, testCommand::getTestState)).withPosition(0, 0);
+    tab.add("Start Testing", new TestCommand(intakeSubsystem,
```
agasser: You don't need two instances of `TestCommand`. Your LEDs aren't lighting up because `LEDProgressBarCommand` is calling a different instance of `TestCommand` than the one on ShuffleBoard. Just use `testCommand` here instead of a new instance and the LEDs will light up.

agasser (changes requested):
With these changes, I think we're ready to try it on the robot.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -276,7 +279,13 @@ public void populateSysIdDashboard() {
...
     tab.add("Tur Roll Dynam Fwd", turretSubsystem.sysIdRollerDynamicCommand(kForward)).withPosition(columnIndex, 2);
     tab.add("Tur Roll Dynam Rev", turretSubsystem.sysIdRollerDynamicCommand(kReverse)).withPosition(columnIndex, 3);
-
+  
+    // Testing
+    tab = Shuffleboard.getTab("Testing");
+    
+    tab.addNumber("Number of Tests Run", () -> testCommand.getTestState());
+    tab.add("Start LED Progress Bar", new LEDProgressBarCommand(ledSubsystem, testCommand::getTestState)).withPosition(0, 0);
+    tab.add("Start Testing", testCommand);
```
agasser: I have one final improvement to suggest. You could run the LEDs and tests in a deadline group:
```
    tab.add("Start Testing", testCommand
        .deadlineWith(new LEDProgressBarCommand(ledSubsystem, testCommand::getTestState)).withName("Test Mode"));
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -284,8 +284,9 @@ public void populateSysIdDashboard() {
     tab = Shuffleboard.getTab("Testing");
     
     tab.addNumber("Number of Tests Run", () -> testCommand.getTestState());
-    tab.add("Start LED Progress Bar", new LEDProgressBarCommand(ledSubsystem, testCommand::getTestState)).withPosition(0, 0);
-    tab.add("Start Testing", testCommand);
+    tab.add("Start Testing", testCommand
+    .deadlineWith(new LEDProgressBarCommand(ledSubsystem, testCommand::getTestState)).withName("Test Mode"));
```
agasser, line 288: Indenting is wrong

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
```diff
@@ -14,7 +14,7 @@ public class LEDProgressBarCommand extends Command {
   private final IntSupplier intSupplier;
   private final LEDSubsystem ledSubsystem;
   private int index;
-  private int tests;
+  private double tests;
```
agasser, line 17: This should be an `int`, it's a whole number.

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
```diff
@@ -30,16 +30,34 @@ public void initialize() {
...
+              ledStrips.setLED(strip, index, Color.kPink);
+            }
+          } else {
+            ledStrips.setLED(strip, index, Color.kBlack);
+          }
         }
       }
+      ledStrips.refresh();
+    } else {
+      ledStrips.setAll(Color.kLimeGreen);
```
agasser, line 59: `kLimeGreen` was used twice. Should the colors be unique?

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,77 @@
...
+import edu.wpi.first.wpilibj2.command.Command;
+import frc.robot.subsystems.LEDStrips;
+import frc.robot.subsystems.LEDSubsystem;
+
+/** Command to run the boot animation on the LED strips */
+public class LEDProgressBarCommand extends Command {
+
+  private final IntSupplier intSupplier;
+  private final LEDSubsystem ledSubsystem;
+  private int index;
```
agasser, line 16: Why is index declared here instead of in the `for` loop?

agasser (changes requested):

PR #12, opened 2024-03-13
Fixed Shoot while move calculation error

dylanrattai:
Time = Coefficient * (Distance / Velocity) 
not
Time = Coefficient * Distance * Velocity

Credits to Adam from CIS for noticing this error in my formula.


ItsMajestiX:
Thanks! You should make the same change in the [auto version](https://github.com/STMARobotics/frc-7028-2024/blob/24b9d5bcf9b65bf69e9cf4c11c46e827b5fa9d13/src/main/java/frc/robot/commands/ScoreSpeakerAutoCommand.java#L88) of this command. You'll also need to change `SHOOT_WHILE_MOVING_COEFFICIENT`. If your robot is like ours, start looking around 300.

agasser (approved):

PR #13, opened 2024-04-03
Test mode

finnsherman:
Test mode for Tempo


src/main/java/frc/robot/commands/led/DefaultLEDCommand.java
```diff
@@ -75,8 +81,10 @@ private LEDMode getMode() {
       return LEDMode.MODE_NOTE_IN_TURRET;
     } else if (!DriverStation.isDSAttached()) {
       return LEDMode.MODE_DS_DISCONNECT;
-    } else if (RobotState.isDisabled()) {
+    } else if (RobotState.isDisabled() && !RobotState.isTest()) {
```
agasser: If you flip these around, you don't need the &&:
```
} else if (RobotState.isTest()) {
 ...
} else if (RobotState.isDisabled()) {
...
}

File: src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java [hunk 2]
```diff
@@ -0,0 +1,98 @@
...
+    this.intSupplier = intSupplier;
+
+    addRequirements(ledSubsystem);
+  }
+
+  @Override
+  public void initialize() {
+    timer.stop();
+    timer.reset();
+
```
agasser, line 37: You should reset your state variables, `tests`, `ledsOn`, and `hasBlinked`, in `initialize()`

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -266,8 +270,16 @@ public void populateSysIdDashboard() {
...
+    tab.add("Tur Roll Dynam Rev", turretSubsystem.sysIdRollerDynamicCommand(kReverse)).withPosition(columnIndex, 3);    
+  }
 
+  public void populateTestingDashboard() {
+    var tab = Shuffleboard.getTab("Testing");
+    tab.addNumber("Number of Tests Run", () -> testCommand.getTestState());
+    tab.addBoolean("Has Stopped", () -> testCommand.getHasStopped());
+    tab.add("Start Testing", testCommand);
+    tab.add("lights", new LEDProgressBarCommand(ledSubsystem, testCommand::getTestState));
+    tab.add("add test state", new InstantCommand(testCommand::addTestState));
```
agasser: This doesn't seem right, why are there three commands going on the dashboard for test mode. I think this should only be one [command group](https://docs.wpilib.org/en/2021/docs/software/commandbased/command-groups.html).

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,98 @@
...
+              ledStrips.setLED(strip, index, Color.kLimeGreen);
+            } else if (index < 15) {
+              ledStrips.setLED(strip, index, Color.kBlue);
+            } else if (index < 18) {
+              ledStrips.setLED(strip, index, Color.kPurple);
+            } else if (index < 21) {
+              ledStrips.setLED(strip, index, Color.kPink);
+            } else if (index < 24) {
+              ledStrips.setLED(strip, index, Color.kWhite);
+            }
```
agasser, line 64: Not all of the strips are the same length, so this logic doesn't look right. The back strip is only 18 LEDs and the sides are 20.
I think it might be better off using `setLEDSegments`, this is what it's for. It just lights up the side strips, and it does all the math automatically to determine how many LEDs to light up in each segment.

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        hasBlinked++;
+      }
+
+    }
+  }
+
+  @Override
+  public void execute() {
+    tests = (intSupplier.getAsInt()*3);
```
agasser, line 89: What is the "magic number" 3?

src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -0,0 +1,208 @@
...
+    this.shooterSubsystem = shooterSubsystem;
+    this.turretSubsystem = turretSubsystem;
+
+    addRequirements(intakeSubsystem, shooterSubsystem, turretSubsystem);
+  }
+
+  @Override
+  public void initialize() {
+    timer.reset();
+    teststate = 0;
```
agasser, line 38: You should reset all of your state variable in `initialize()`, you missed `hasStopped`.

src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -0,0 +1,208 @@
...
+          timer.start();
+          hasStopped = true;
+        }
+        if (timer.hasElapsed(3)) {
+          teststate++;
+          timer.stop();
+          timer.reset();
+          hasStopped = false;
+        }
+
```
agasser, line 167: You're missing a `break` here, that's why it's [falling through](https://www.geeksforgeeks.org/fall-through-condition-in-java/) to moving the pitch (or setting the yaw to the wrong direction, like we saw last week.)

src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -0,0 +1,208 @@
...
+          turretSubsystem.prepareToIntake();
+          timer.start();
+          hasStopped = true;
+        }
+        if (timer.hasElapsed(3)) {
+          teststate++;
+          timer.stop();
+          timer.reset();
+          hasStopped = false;
+        }
```
agasser, line 182: It's not _needed_. but I would put a `break` here. If you don't, you're at risk that you'll add another `case` and forget it, like in the previous `case`.

src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -0,0 +1,208 @@
...
+      case 0:
+        if (!hasStopped) {
+          intakeSubsystem.runRollers(RotationsPerSecond.of(20));
+          timer.start();
+        }
+        if (timer.hasElapsed(2) && !hasStopped) {
+          intakeSubsystem.stop();
+          hasStopped = true;
+        }
+        if (timer.hasElapsed(4) && hasStopped == true) {
```
agasser: This is OK, but I highly recommend not writing it as `hasStopped == true` because you're at risk of making a simple mistake by forgetting one of the `=` signs. You could easily have `hasStopped = true`, and this will compile, but it will set `hasStopped` to `true`, and always evaluate to `true`.

TL;DR;
Change this to:
```
if (timer.hasElapsed(4) && hasStopped) {
```
or if you want to keep `==`, you should put the literal first so it won't compile if you forget a `=`
```
if (timer.hasElapsed(4) && true == hasStopped) {
```

src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  @Override
+  public void initialize() {
+    timer.reset();
+    teststate = 0;
+  }
+
+  @Override
+  public void execute() {
```
agasser, line 43: Just a suggestion, but the logic could be simplified if you didn't combine forward, stop, and reverse all in the same state. If you advanced the state every 2 seconds instead, you could get rid of `hasStopped`. Something like this:
```java
    timer.start();
    switch (teststate) {
      case 0:
        intakeSubsystem.runRollers(RotationsPerSecond.of(20));
        checkTimer();
        break;

      case 1:
        intakeSubsystem.stop();
        checkTimer();
        break;
      
      case 2:
        intakeSubsystem.runRollers(RotationsPerSecond.of(-20));
        checkTimer();
        break;
...

  private void checkTimer() {
    if (timer.hasElapsed(2)) {
      timer.reset();
      teststate++;
    }
  }
```

Or make it a sequential command group.

agasser (changes requested):

src/main/java/frc/robot/commands/testing/LEDProgressBarCommand.java
finnsherman: It's used to make 3 LEDs light up per test done

PR #17, opened 2024-06-17
One babybird

finnsherman:
add a command where the driver holds down one button and the robot automatically intakes the notes from the human player and shoots them to the middle.

### DIFF
--- a/src/main/java/frc/robot/AutoCommands.java
+++ b/src/main/java/frc/robot/AutoCommands.java
@@ -172,7 +172,8 @@ public class AutoCommands {
   public Command shootMid() {
     return new ShootTeleopCommand(
           drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, desiredX,
-          desiredY, STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3);
+          desiredY, STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
+          .deadlineWith((new LEDAlternateCommand(ledSubsystem, Color.kBlue, Color.kOrange, Seconds.of(0.1))));
   }
 
 }
--- a/src/main/java/frc/robot/AutoCommands.java
+++ b/src/main/java/frc/robot/AutoCommands.java
@@ -9,9 +9,14 @@ import static frc.robot.Constants.ShootingConstants.STOCKPILE_INTERPOLATOR;
 import static frc.robot.Constants.ShootingConstants.STOCKPILE_MID_BLUE;
 import static frc.robot.Constants.ShootingConstants.STOCKPILE_MID_RED;
 
+import java.util.function.Supplier;
+
 import com.pathplanner.lib.auto.NamedCommands;
 
 import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.units.Distance;
+import edu.wpi.first.units.Measure;
+import edu.wpi.first.units.Velocity;
 import edu.wpi.first.wpilibj.util.Color;
 import edu.wpi.first.wpilibj2.command.Command;
 import frc.robot.Constants.LEDConstants;
@@ -25,8 +30,6 @@ import frc.robot.commands.ShootTeleopCommand;
 import frc.robot.commands.led.LEDAlternateCommand;
 import frc.robot.commands.led.LEDBlinkCommand;
 import frc.robot.commands.led.LEDMarqueeCommand;
-import frc.robot.controls.ControlBindings;
-import frc.robot.controls.JoystickControlBindings;
 import frc.robot.subsystems.CommandSwerveDrivetrain;
 import frc.robot.subsystems.IntakeSubsystem;
 import frc.robot.subsystems.LEDSubsystem;
@@ -45,8 +48,6 @@ public class AutoCommands {
   private final TurretSubsystem turretSubsystem;
   private final IntakeSubsystem intakeSubsystem;
   private final LEDSubsystem ledSubsystem;
-
-  private final ControlBindings controlBindings = new JoystickControlBindings();
 
   public AutoCommands(CommandSwerveDrivetrain drivetrainSubsystem, ShooterSubsystem shooterSubsystem,
       TurretSubsystem turretSubsystem, IntakeSubsystem intakeSubsystem, LEDSubsystem ledSubsystem) {
@@ -156,16 +157,19 @@ public class AutoCommands {
         turretSubsystem, shooterSubsystem).deadlineWith(new LEDBlinkCommand(ledSubsystem, BABYBIRD_COLOR, 0.1));
   }
 
-  /**
+    /**
    * Comand to shoot a note into the middle of the field
+   * @param xSupplier X translation supplier
+   * @param ySupplier Y translation supplier
    * @return new command
    */
 
-  public Command shootMid() {
+  public Command shootMid(
+      Supplier<Measure<Velocity<Distance>>> xSupplier, Supplier<Measure<Velocity<Distance>>> ySupplier) {
     return new ShootTeleopCommand(
-          drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, controlBindings.translationX(),
-          controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
-          .deadlineWith((new LEDAlternateCommand(ledSubsystem, Color.kBlue, Color.kOrange, Seconds.of(0.1))));
+        drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, xSupplier,
+        ySupplier, STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
+            .deadlineWith((new LEDAlternateCommand(ledSubsystem, Color.kBlue, Color.kOrange, Seconds.of(0.1))));
   }
 
 }
--- a/src/main/java/frc/robot/AutoCommands.java
+++ b/src/main/java/frc/robot/AutoCommands.java
@@ -3,7 +3,7 @@ package frc.robot;
 import static edu.wpi.first.units.Units.Degrees;
 import static edu.wpi.first.units.Units.RotationsPerSecond;
 import static edu.wpi.first.units.Units.Seconds;
-import static frc.robot.Constants.LEDConstants.FLASH_COLOR;
+import static frc.robot.Constants.LEDConstants.BABYBIRD_COLOR;
 import static frc.robot.Constants.LEDConstants.NOTE_COLOR;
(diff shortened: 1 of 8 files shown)


src/main/java/frc/robot/Constants.java
```diff
@@ -273,7 +273,7 @@ public static class LEDConstants {
     public static final int DEVICE_ID = 9;
 
     public static final Color NOTE_COLOR = Color.fromHSV(3, 255, 255);
-
+    public static final Color FLASH_COLOR = Color.fromHSV(255, 255, 255);
```
agasser: Why is the LED color different when doing baby-bird in this mode?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -191,6 +195,14 @@ private void configureButtonBindings() {
         drivetrain, shooterSubsystem, turretSubsystem, ledSubsystem, controlBindings.translationX(),
         controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)));
     
+    controlBindings.babyBomber().ifPresent(trigger -> trigger.whileTrue(
+      new BabyBirdCommand(
+        turretSubsystem, shooterSubsystem).deadlineWith(new LEDBlinkCommand(ledSubsystem, FLASH_COLOR, 0.1))
+      .andThen(new ShootTeleopCommand(
+        drivetrain, shooterSubsystem, turretSubsystem, ledSubsystem, controlBindings.translationX(),
+        controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
+        .deadlineWith(new LEDAlternateCommand(ledSubsystem, Color.kBlue, Color.kOrange, Seconds.of(0.1))))));
```
agasser: Indenting is incorrect

agasser: This is OK, but my preference would be to move the construction of `BabyBirdCommand` and `ShootTeleopCommand` into the `AutoCommands` class and then reuse them for both bindings instead of doing copy-paste.

agasser: To match your description of the driver holding the button down and it shooting multiple notes, you'll need to add `.repeatedly()` to the end. `whileTrue` does not reschedule the command when it ends.

agasser (changes requested):
This looks pretty good, you're on the right track!

src/main/java/frc/robot/Constants.java
finnsherman: To signify to the human player that the robot is baby-birding and needs a note

agasser: Blinking with `NOTE_COLOR` already signifies that the robot is in baby-bird mode and the human player should load a note. Why is `FLASH_COLOR` a different color?
I'm not understanding why we flash orange for the human player when baby-birding by itself, but blinking white when baby-birding to stockpile.

finnsherman: We definitely could, Osowski and Scott told me that we'd want the brightest we can achieve to notify the operator that the robot needs the note. There's no issue with changing it to orange though.

agasser: I'm OK changing the color, my question is why is it _different_ from regular baby bird. If you want to use white, that's fine but give the constant a meaningful name (like `BABY_BIRD_COLOR`) and use it for _both_ ways the driver can trigger baby bird. You left the old one using `NOTE_COLOR`.

src/main/java/frc/robot/RobotContainer.java
finnsherman: I just pushed a commit that moves them, however I'm unsure on how to resolve the error as the autocommands object is created before controlbindings is initialized in the constructor.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -196,12 +197,8 @@ private void configureButtonBindings() {
         controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)));
     
     controlBindings.babyBomber().ifPresent(trigger -> trigger.whileTrue(
-      new BabyBirdCommand(
-        turretSubsystem, shooterSubsystem).deadlineWith(new LEDBlinkCommand(ledSubsystem, FLASH_COLOR, 0.1))
-      .andThen(new ShootTeleopCommand(
-        drivetrain, shooterSubsystem, turretSubsystem, ledSubsystem, controlBindings.translationX(),
-        controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
-        .deadlineWith(new LEDAlternateCommand(ledSubsystem, Color.kBlue, Color.kOrange, Seconds.of(0.1))))));
+      autoCommands.babyBird().deadlineWith(new LEDBlinkCommand(ledSubsystem, FLASH_COLOR, 0.5)).andThen(autoCommands.shootMid()
```
agasser: You have the baby bird `LEDBlinkCommand` in `autoCommands.babyBird()` and here, that won't work. I think you should only have it in `autoCommands.babyBird()`.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -170,7 +174,7 @@ private void configureButtonBindings() {
     
     controlBindings.babyBird().ifPresent(trigger -> trigger.whileTrue(
```
agasser: Let's use `autoCommands.babyBird()` here. The reason for moving it into `autoCommands` was so we can reuse it.

src/main/java/frc/robot/AutoCommands.java
```diff
@@ -135,4 +154,25 @@ public Command scoreAmp() {
...
+
+  /**
+   * Comand to shoot a note into the middle of the field
+   * @return new command
+   */
+
+  public Command shootMid() {
+    return new ShootTeleopCommand(
+          drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, desiredX,
+          desiredY, STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3);
```
agasser: I like how you also put the `LEDBlinkCommand` into `babyBird()`, let's do the same here with the `LEDAlternateCommand`.

agasser (changes requested):

src/main/java/frc/robot/RobotContainer.java
agasser: I think a good way to handle this would be to have the control bindings passed to the `shootMid()` method instead of the constructor. This should work well since a new command instance is created each time the method is called.

src/main/java/frc/robot/AutoCommands.java
```diff
@@ -171,8 +163,8 @@ public Command babyBird() {
 
   public Command shootMid() {
     return new ShootTeleopCommand(
-          drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, desiredX,
-          desiredY, STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
+          drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, controlBindings.translationX(),
+          controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
```
agasser: Don't create a new instance of ControlBindings, just take the suppliers as parameters to the method.

```java
  /**
   * Comand to shoot a note into the middle of the field
   * @param xSupplier X translation supplier
   * @param ySupplier Y translation supplier
   * @return new command
   */
  public Command shootMid(
      Supplier<Measure<Velocity<Distance>>> xSupplier, Supplier<Measure<Velocity<Distance>>> ySupplier) {
    return new ShootTeleopCommand(
        drivetrainSubsystem, shooterSubsystem, turretSubsystem, ledSubsystem, xSupplier,
        ySupplier, STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)
            .deadlineWith((new LEDAlternateCommand(ledSubsystem, Color.kBlue, Color.kOrange, Seconds.of(0.1))));
  }
```
Then you can use the existing controlBindings in RobotContainer:
```java
    controlBindings.babyBomber().ifPresent(trigger -> trigger.whileTrue(
        autoCommands.babyBird()
            .andThen(autoCommands.shootMid(controlBindings.translationX(), controlBindings.translationY()))
            .repeatedly()));
```

agasser (changes requested):

agasser:
I just noticed that `ShootTeleopCommand` doesn't have an end condition, so holding the button down will still only run the cycle one time. If you want it to run repeatedly you'll also have to make `ShootTeleopCommand` end when shooting is done. You can probably just copy the way `ScoreSpeakerAutoCommand` does it, but there might be some nuisance I'm not thinking of.

```java
  @Override
  public boolean isFinished() {
    return !turretSubsystem.hasNote();
  }
```

src/main/java/frc/robot/AutoCommands.java
```diff
@@ -135,4 +147,29 @@ public Command scoreAmp() {
...
+   * Command to put the turret in intake position
+   * @return new command
+   */
+
+  public Command babyBird() {
+    return new BabyBirdCommand(
+        turretSubsystem, shooterSubsystem).deadlineWith(new LEDBlinkCommand(ledSubsystem, BABYBIRD_COLOR, 0.1));
+  }
+
+    /**
```
agasser: Indenting is wrong here

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -191,6 +186,11 @@ private void configureButtonBindings() {
         drivetrain, shooterSubsystem, turretSubsystem, ledSubsystem, controlBindings.translationX(),
         controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)));
     
+        controlBindings.babyBomber().ifPresent(trigger -> trigger.whileTrue(
+          autoCommands.babyBird()
+              .andThen(autoCommands.shootMid(controlBindings.translationX(), controlBindings.translationY()))
+              .repeatedly()));
+
```
agasser: This is indented wrong

src/main/java/frc/robot/AutoCommands.java
```diff
@@ -135,4 +147,29 @@ public Command scoreAmp() {
         .deadlineWith(new LEDMarqueeCommand(ledSubsystem, 70, 255, 0, 15, 0.08));
   }
 
+  /**
+   * Command to put the turret in intake position
+   * @return new command
+   */
+
```
agasser: There's an extra line here

agasser (changes requested):
Looks good, except for some formatting issues

agasser:
Nice work. Let's try it out next week!

agasser (approved):

PR #18, opened 2024-06-17
Test mode upgrade

finnsherman:
test mode for IRI


src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -0,0 +1,207 @@
...
+  public boolean isFinished() {
+    return !RobotState.isTest();
+  }
+
+  @Override
+  public void end(boolean interrupted) {
+    shooterSubsystem.stop();
+    turretSubsystem.stop();
+    intakeSubsystem.stop();
+    turretSubsystem.prepareToIntake();
```
agasser: Why `prepareToIntake()` at the end? Usually you would just want to stop the subsystems when you're done with them.

agasser (changes requested):
Really close, just one question...

src/main/java/frc/robot/commands/testing/TestCommand.java
finnsherman: No idea what I was thinking when I wrote that, but its fixed now.

src/main/java/frc/robot/commands/testing/TestCommand.java
```diff
@@ -202,6 +202,8 @@ public void end(boolean interrupted) {
     shooterSubsystem.stop();
     turretSubsystem.stop();
     intakeSubsystem.stop();
-    turretSubsystem.prepareToIntake();
+    turretSubsystem.stopPitch();
+    turretSubsystem.stopYaw();
+    turretSubsystem.runRollers(RotationsPerSecond.zero());
```
agasser: Line 203 calls `turretSubsystem.stop()`, so I don't think you need lines 205-207, the turret will already be stopped.

(Side note, `runRollers(RotationsPerSecond.zero())` doesn't turn the rollers off, it actively drives and holds the velocity at zero. `stopRollers()` is the method to turn the roller motors off.)

agasser (changes requested):

agasser (approved):
Test it on the robot, and then you're good to merge it!

PR #19, opened 2024-07-10
changes from 7/10/2024

finnsherman:


src/main/java/frc/robot/Constants.java
```diff
@@ -117,7 +117,7 @@ public static class VisionConstants {
      * Minimum target ambiguity. Targets with higher ambiguity will be discarded. Not appliable when multiple tags are
      * in view in a single camera.
      */
-    public static final double APRILTAG_AMBIGUITY_THRESHOLD = 0.2;
+    public static final double APRILTAG_AMBIGUITY_THRESHOLD = 0.15;
```
agasser: Please explain why we want this change

finnsherman: The left camera on the robot was struggling to maintain its position when it only saw one apriltag, so we adjusted the ambiguity ratio to reduce the amount the robot was shuffling its position. (We had already tried adjusting camera brightness and exposure beforehand)

agasser: It's your call, but I recommend changing it back to `0.2`, and here's why:
- If the position was not stable when sitting still and seeing only one tag, the threshold is _too low_ not _too high_. This would indicate pose estimation is using tags that are _too_ ambiguous and you aren't getting a consistent reading between frames. Those readings show be thrown out.
- `0.2` is the value we used for every event for the last two seasons, including Champs. It's battle tested at multiple events on a competition field.

agasser (approved):

PR #21, opened 2024-08-15
Faster babybomber

finnsherman:
I wanted something to do and remembered a comment Osowski made at IRI about how baby bomber would be faster if it didn't reset the turret's position every time so I went ahead and did that. I'm pretty sure it will work but it will obviously need to be tested sometime during preseason. I really do not need to be there for the testing so if anyone gets the chance feel free to test it and leave a comment or merge the branch if it works, and request any changes or report any malfunctions if it does not do what it's supposed to.


agasser:
I don't think you need to copy the commands into a single command. You probably just need to remove/override the `end()` method on `BabyBirdCommand`, I think stopping the motors is what's making the turret fall.

The simplest, but a little hacky, way to do it is with an anonymous inner class:
```java
  public Command babyBirdNoDrop() {
    return new BabyBirdCommand(
        turretSubsystem, shooterSubsystem) {
          public void end(boolean interrupted) {
            // Don't turn the motors off
          };
        }.deadlineWith(new LEDBlinkCommand(ledSubsystem, BABYBIRD_COLOR, 0.1));
  }
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -190,9 +190,9 @@ private void configureButtonBindings() {
         controlBindings.translationY(), STOCKPILE_MID_RED, STOCKPILE_MID_BLUE, STOCKPILE_INTERPOLATOR, 0.3)));
     
     controlBindings.babyBomber().ifPresent(trigger -> trigger.whileTrue(
-        autoCommands.babyBird()
-          .andThen(autoCommands.shootMid(controlBindings.translationX(), controlBindings.translationY()))
-          .repeatedly()));
+      autoCommands.babyBirdNoDrop()
+      .andThen(autoCommands.shootMid(controlBindings.translationX(), controlBindings.translationY()))
+      .repeatedly()));
```
agasser: These lines are not indented correctly

agasser (changes requested):
I think this looks right (except the indenting). Let's give it a try when we can get on the robot!

agasser (commented):
Looks great. We need to test before approving and merging.

PR #22, opened 2024-08-26
Flashing red LEDs when the camera is disconnected from the robot

finnsherman:
based on [this post](https://7028robotics.slack.com/archives/C7CA91QKU/p1724539778064669) in slack

### DIFF
--- a/src/main/java/frc/robot/PhotonRunnable.java
+++ b/src/main/java/frc/robot/PhotonRunnable.java
@@ -54,6 +54,8 @@ public class PhotonRunnable implements Runnable {
 
   private final Supplier<Pose2d> poseSupplier;
 
+  public boolean isCameraConnected;
+
   @SuppressWarnings("unchecked")
   private final StructArrayPublisher<AprilTag>[] aprilTagPublishers = new StructArrayPublisher[2];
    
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/PhotonRunnable.java
```diff
@@ -54,6 +54,8 @@ public class PhotonRunnable implements Runnable {
 
   private final Supplier<Pose2d> poseSupplier;
 
+  public boolean isCameraConnected;
```
agasser, line 57: This field shouldn't be `public`. By making it `public`, its value can be modified outside of the class. You already have a getter/accessor, so just make it `private`.

src/main/java/frc/robot/PhotonRunnable.java
```diff
@@ -100,6 +102,17 @@ public PhotonRunnable(String[] cameraNames, Transform3d[] robotToCameras, AddVis
       photonPoseEstimators[i] = new PhotonPoseEstimator(
           layout, MULTI_TAG_PNP_ON_COPROCESSOR, new PhotonCamera(cameraNames[i]), robotToCameras[i]);
     }
+    final PhotonCamera leftCamera = new PhotonCamera("Left");
+    final PhotonCamera rightCamera = new PhotonCamera("Right");
+    if (leftCamera.isConnected() && rightCamera.isConnected()) {
+      isCameraConnected = true;
+      leftCamera.close();
+      rightCamera.close();
+    }
```
agasser, line 111: We need to do a little thinking on how to detect if the cameras are connected. What you have would work, but only if the cameras are connected when `PhotonRunnable` is instantiated. It only runs once, so if they come online later it wouldn't detect it. It would also fail to detect if the cameras were connected, but then disconnected later.

We probably need to have a timeout where we consider a camera disconnected if ~it hasn't sent~ we haven't received any data for a while, maybe one second. We have to do something within the `run()` method, this is the thread's code that runs continuously. One consideration is that the thread blocks (waits) at `WPIUtilJNI.waitForObjects()` if there is no new camera data, so the thread can tell you when the cameras _are_ connected, but something else will have to watch for when we stop receiving data.

PR #23, opened 2025-04-25
change demo mode to xbox controller

finnsherman:


agasser (approved):
Look good. I don't know if we want to merge this to `main`, but I'm cool with it.

